PR #11, opened 2018-06-20
 Extract out a concise Portals v0 summary from the explainer.

jeremyroman:
In the hopes of coming up with a crisp and concise v0 proposal, this PR extracts
out detailed design discussion and most mentions of privacy-enhancing modes into
separate documents. It's valuable, but makes the explainer itself somewhat intimidating.

This leaves a short explanation of how a portal would facilitate seamless navigation as
a v0 target. If we flesh out followup steps for privacy or performance isolation features,
they can go in separate documents which expand on the first.


jeremyroman:
ping @KenjiBaheux ?

jeremyroman:
With no feedback, going to merge.

Issue #13, opened 2018-06-28
Reflect takeaways from talking to Legal/Privacy about GDPR


KenjiBaheux:
Need to update the design-discussion document.

PR #14, opened 2018-06-28
GDPR considerations

KenjiBaheux:
See issue #13


KenjiBaheux:
@jeremyroman just an FYI, in case I'm missing something.

PR #24, opened 2018-09-17
Spec draft

jeremyroman:
This is a first pass at formalizing the portals work into a specification. There's still some stuff left out here to avoid making this PR any bigger. It's largely noted either in inline issues in the spec or existing GitHub issues.

Preview: https://rawgit.com/jeremyroman/portals/spec-draft/index.html

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -234,7 +234,7 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
   </section>
 
   <section algorithm="htmlportalelement-acceptpostedmessage">
-    To <dfn for="HTMLPortalElement">[=accept a message posted to the host=] for a <{portal}> element</dfn> with
+    To [=accept a message posted to the host=] for a <{portal}> element with
     |serializeWithTransferResult|, |origin| and |targetOrigin|, [=queue a task=] from the
     [=posted message task source=] to the [=event loop=] associated with the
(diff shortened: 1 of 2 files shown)


jeremyroman:
@lucasgadani I can't seem to assign you as a reviewer, but I'd also like you to look at this.

I'm sure we'll have to get Domenic (or someone else) to take a proper look at this, but since we're still early days it's probably fine to stage this in the GitHub repo and do a review from there. WDYT?

lucasgadani:
Thanks for working on this! The draft so far LGTM.

I think we should expand a bit on the lifetime of the PortalHost object, and when it has an associated browsing context. In particular, since we are dealing with two independent event loops, the portal browsing context's event loop needs to be notified that it's PortalHost object is no longer associated host (i.e. this isn't a synchronous operation as it might be implied in the "activate a portal browsing context" in section 2).

KenjiBaheux:
@lucasgadani I've added you to the contributors, check your inbox for an invite link.

jeremyroman:
PTAL.

lucasgadani:
Thanks, the draft LGTM.

I think we'll also need to think about and include the fallback behavior of <portal>. In particular, what will happen with children of the <portal> element, but it should be ok merging this and working on this as a follow-up.

jeremyroman:
Yes, I don't mean to imply that there are no remaining things to describe. :smile:

PR #31, opened 2018-11-20
Allow passing structured cloneable data to HTMLPortalElement/activate

### LINKED ISSUE #28: Feature request: passing data with activate()
When writing some demo code with portals, we've found it to be very common to need to pass data to a portal "while" activating it.

The ideal syntax for this would be

```js
// outside.js
portal.activate({ some: 'structured cloneable data' });

// inside.js
window.addEventListener("portalactivate", e => {
  console.log(e.data);
});
```

Right now I am emulating it like so: https://gist.github.com/domenic/36cc626e1ed748d26afe6c83ca732ca0 (although I haven't tested that this works as Chrome Canary doesn't have a working portalactivate event yet).

jeremyroman:
Uses `StructuredSerialize`, which doesn't allow transfer but is simple.

Resolves #28.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -80,7 +80,8 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
 
   <section algorithm="portal-browsing-context-activate">
     To <dfn>activate a portal browsing context</dfn> |portalBrowsingContext| in
-    place of |predecessorBrowsingContext|, run the following steps [=in parallel=]:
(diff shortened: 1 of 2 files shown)


jeremyroman:
Updated to use `StructuredSerializeWithTransfer` as discussed in #28.

jeremyroman:
Preview: https://rawgit.com/jeremyroman/portals/activate-data/

PR #34, opened 2018-12-05
Fix w3c.json syntax.

jeremyroman:
Per https://w3c.github.io/w3c.json.html, the group value should be a number or array of numbers, and the repo-type value should be a quoted string.

### DIFF
--- a/w3c.json
+++ b/w3c.json
@@ -1,5 +1,5 @@
  {
-    "group":      ["80485"]
+    "group":      80485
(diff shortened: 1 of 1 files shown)


yoavweiss:
Hmm, that was automatically generated by https://github.com/w3c/ash-nazg so that's probably exposing an issue there...

yoavweiss:
Marked as non substantive for IPR from ash-nazg.

PR #39, opened 2018-12-10
Update self-links to point at WICG/portals.

jeremyroman:
These links should be updated to reflect the repository move. LGTY?


lucasgadani:
LGTM

PR #45, opened 2019-01-08
Disallow creating a portal browsing context inside a nested browsing context.

jeremyroman:
We should not allow such things to be activated (if you're in an iframe, you cannot take over the main frame), and the simplest way to do so is to not allow them to exist in the first place. If we later change our mind it would be easy enough to move this check to activation. This is the more conservative approach.

It is easy enough for a page to know that it is in a nested browsing context (checking `window.parent`), and there's unfortunately not a natural way to report this error in the API, so I've included non-normative text suggesting a developer warning.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -110,6 +110,8 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
 
         1. Set |event|'s [=PortalActivateEvent/predecessor browsing context=] to |predecessorBrowsingContext|.
 
+        1. Set |event|'s [=PortalActivateEvent/successor window=] to |successorWindow|.
+
         1. [=Dispatch=] |event| to |successorWindow|.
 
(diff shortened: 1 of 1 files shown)


lucasgadani:
This make sense and LGTM, but it should also be specified  when adopting the predecessor, otherwise a nested browsing context could adopt the predecessor's browsing context.

jeremyroman:
Good point. I'd thought that wasn't possible because the `portalactivate` event is only delivered to portal contexts, but it could pass another document. @domenic had previously suggested inferring the `Document` argument anyhow, and that seems like it would resolve this. WDYT?

lucasgadani:
Inferring the document SGTM for adopting the predecessor.

As we discussed offline, we would also need to drop the portal browsing context when adopting (as in adoptNode) the HTMLPortalElement.

jeremyroman:
PTAL

PR #47, opened 2019-01-10
Clarify the preconditions around postMessage between a portal contents and its host.

jeremyroman:
The question of what the host and guest event loops know about each other during `postMessage` came up when talking with @a4sriniv. This is one take, but review closely as I'm not certain whether these are the ideal semantics.


jeremyroman:
Also a spelling fix. :)

PR #48, opened 2019-01-10
Annotate portals spec with links to corresponding WPT.

jeremyroman:
This is an interesting feature that Bikeshed has that might prove useful. You can annotate the spec sections with the corresponding tests and so identify sections of the spec that aren't well covered (or so the theory goes). I've put it in inline display mode (where it actually renders a list of the tests with links to the source and a live version), but it can also be put in a mode where it doesn't render anything.

It does, however, generate errors if there are WPT tests that aren't listed in the spec (the idea being that the spec editors should be update for new WPT tests shortly thereafter). I'm not sure how I feel about that, but maybe it's worth a shot.


jeremyroman:
Yeah, will tweak it if it ends up being a headache.

PR #58, opened 2019-02-14
Reject activation of nested portals and explicitly delineate the 'orphaned' state during activation.

### LINKED ISSUE #57: Reject activation of nested portals
As we discussed in person, there is considerable complexity that arises if activation of portals within portals is allowed, especially around the case where a context activates a portal while it is itself a portal browsing context undergoing activation. There are also tricky cases around the guarantees that might be desirable regarding queued messages.

You suggested that we simply make it an error to activate a portal from a portal browsing context (i.e., the predecessor context must not be a portal browsing context). It seems to me that this should suffice for the use cases we currently envision and doesn't seem like it will unnecessarily restrict us from enabling this in the future.

The only downside I can see is that portal browsing contexts won't be able to use portals to create cross-origin navigations within themselves, but that seems acceptable to me. Especially since there's typically no browser chrome for embedded content, you could indistinguishably fill the viewport with the portal and wait to activate it until the outer portal is activated, with only a small loss of fidelity.

Does making this change SGTY?

jeremyroman:
Reject activation of nested portals and explicitly delineate the 'orphaned' state during activation.

Some minor changes for consistency.

Resolves #57.


index.bs
```diff
@@ -229,6 +254,9 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
 
         If no such context exists, throw an "{{InvalidStateError}}" {{DOMException}}.
 
+    1. If the [=portal state=] of |predecessorBrowsingContext| is not *none*,
+        throw an "{{InvalidStateError}}" {{DOMException}}.
```
lucasgadani: Let's add some non-normative text that this means that nested portals can't be activated.

PR #90, opened 2019-04-15
Modernize event construction and include required [Constructor].

### LINKED ISSUE #84: PortalActivateEvent IDL
Needs a [Constructor], like all modern event classes. It will get auto-generated by https://dom.spec.whatwg.org/#constructing-events, although you should probably define some "event constructing steps" to set predecessor BC and successor window to null.

Needs [Exposed=Window].

jeremyroman:
Resolves #84.


jeremyroman:
FYI @domenic.

PR #91, opened 2019-04-16
Remove use of 'context object' from dom.spec.whatwg.org.

### LINKED ISSUE #75: "context object" is going away
We finally added a first-class notion of "this": https://heycam.github.io/webidl/#this

You can reference it with `<b>this</b>`; first example usage at https://github.com/whatwg/encoding/pull/175. It should flow through everything eventually.

jeremyroman:
Resolves #75.


jeremyroman:
attn @domenic

PR #95, opened 2019-04-16
Link to global object realm when accessing via a Window object.

### LINKED ISSUE #64: "Let targetRealm be successorWindow’s realm" has wrong cross-link
realm links to https://html.spec.whatwg.org/multipage/webappapis.html#environment-settings-object's-realm but it should link to https://html.spec.whatwg.org/multipage/webappapis.html#concept-global-object-realm

jeremyroman:
Resolves #64.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -122,7 +122,7 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
 
         1. Set the [=portal state=] of |successorBrowsingContext| to *none*.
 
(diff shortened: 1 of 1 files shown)


jeremyroman:
@domenic ptal

Also, despite adding the `.pr-preview.json` file, I don't seem to be getting PR previews. :(

jeremyroman:
Yeah, that matches what I see. :/

PR #96, opened 2019-04-16
Fix usage of StructuredDeserializeWithTransfer.

### LINKED ISSUE #66: StructuredDeserializeWithTransfer usage tweak
> Initialize event’s data attribute to dataClone.

StructuredDeserializeWithTransfer returns a record with two fields, of which you want the [[Deserialized]] field, probably. See https://html.spec.whatwg.org/#window-post-message-steps for some example usage, including variable names.

This may have changed a while back, between when you wrote the spec and now.

jeremyroman:
Resolves #66.


jeremyroman:
@domenic

PR #100, opened 2019-04-18
Adopt the spec concept of 'browsing-context connected' to simplify insertion steps.

### LINKED ISSUE #80: "Whenever a portal element is inserted"
Portals should probably work in shadow DOM, I think. (Iframes should as well, but apparently aren't specced to do so????).

<del>Portals should probably not work inside non-browsing-context documents, I think.</del> <ins>I see this is already covered by the later steps of the algorithm, so the below is just an editorial suggestion in this regard.</ins>

In that case, you should use https://html.spec.whatwg.org/#becomes-browsing-context-connected

(Same for "removed" steps a bit later.)

Also, there's a "the the" in step 2, and maybe it should cross-reference nested BC.

jeremyroman:
Resolves #80.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -31,6 +31,9 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
             text: refuse to allow the document to be unloaded; url: refused-to-allow-the-document-to-be-unloaded
         urlPrefix: history.html
             text: session history; url: session-history
+        urlPrefix: infrastructure.html
+            text: becomes browsing-context connected; url: becomes-browsing-context-connected
+            text: becomes browsing-context disconnected; url: becomes-browsing-context-disconnected
         urlPrefix: origin.html
             text: origin; url: concept-origin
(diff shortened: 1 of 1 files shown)


jeremyroman:
@domenic

index.bs
```diff
@@ -454,7 +460,7 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
     portals-nested.html
   </wpt>
 
-  Whenever a <{portal}> element |element| is [=removing steps|removed=], run the following steps:
+  Whenever a <{portal}> element |element| [=becomes browsing-context disconnected=], run the following steps:
```
domenic: These will now run in the BC gets discarded (e.g. if you have a Document with a portal, and you .close() its containing Window).

jeremyroman: That sounds reasonable — the guest contents should be discarded in that case.

jeremyroman:
@lucasgadani sgty?

PR #101, opened 2019-04-19
Various minor improvements to description of window.portalHost getter.

### LINKED ISSUE #85: portalHost attribute nits
We've recently fixed up many of the IDL attributes you modeled this on in small ways. Maybe take a pass at aligning with https://html.spec.whatwg.org/#dom-frameelement. For example "x's getter" instead of "x", and the use of "this Window object's browsing context"---which might be null---now that we have that definition.

jeremyroman:
Resolves #85.

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/portals/pull/101.html" title="Last updated on Apr 23, 2019, 8:13 PM UTC (f336965)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/101/4c40482...jeremyroman:f336965.html" title="Last updated on Apr 23, 2019, 8:13 PM UTC (f336965)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -450,7 +450,6 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
   1. [=set the source URL of a portal element|Set the source URL=] of |element|.
 
   <wpt>
-    portals-create-orphaned.html
     portals-nested.html
   </wpt>
 
@@ -659,7 +658,7 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
   <section algorithm="window-portalhost">
     The <dfn attribute for="Window">portalHost</dfn> attribute's getter *must* run the following steps:
 
-    1. Let |context| be [=this=] {{Window}} object's [=browsing context=].
+    1. Let |context| be [=this=]'s [=browsing context=].
 
     1. If |context| is null or the [=portal state=] of |context| is not "`portal`", then return null.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -509,7 +509,7 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
   The `PortalHost` interface {#the-portalhost-interface}
   ------------------------------------------------------
 
-  The <dfn>portal host object</dfn> of a [=browsing context=] is a {{PortalHost}}.
+  The <dfn>portal host object</dfn> of a {{Window}} is a {{PortalHost}}.
```
jeremyroman: Noticed an inconsistency about which object has the portal host object. It should be the Window, since the platform object should be different across navigations, even if the new one's operations will ultimately do the same thing. (Else the thing will leak across realms etc.)

jeremyroman:
@domenic

index.bs
```diff
@@ -657,19 +657,13 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
   </xmp>
 
   <section algorithm="window-portalhost">
-    The <dfn attribute for="Window">portalHost</dfn> attribute *must* run the following steps:
+    The <dfn attribute for="Window">portalHost</dfn> attribute's getter *must* run the following steps:
 
-    1. Let |windowProxy| be [=this=]'s {{WindowProxy}} object.
+    1. Let |context| be [=this=] {{Window}} object's [=browsing context=].
```
domenic: You can just say "this's" if you want. Now that it's formally defined we'll probably eventually update HTML to use it uniformly, instead of "this X object". (But, either is fine.)

jeremyroman: Okay, I'll say "this's". Also fixed the xref to "Window/browsing context" instead of the general definition of "browsing context".

domenic: Doh, I'll try to check linking destinations better next time!

PR #103, opened 2019-04-24
Clarify no-portals-in-subframes note.

### LINKED ISSUE #82: Adopting into other TLBCs?
When a portal element is adopted, it gets closed.

This is immediately followed by a note stating

> Since portal elements can have only a guest browsing context while their document belongs to a top-level browsing context, it is impossible to embed a portal browsing context in a nested browsing context. 

It's not clear whether this note is intended to justify the adopting steps, or is just vaguely related.

So, is it intended that adopting a portal element into a different TLBC (e.g., into a portal BC, or into an auxiliary BC) also closes the portal? If so, maybe the note should be tweaked. Something like

> Since portal elements can have only a guest browsing context while their document belongs to a top-level browsing context, it is impossible to embed a portal browsing context in a nested browsing context. We also close the guest browsing context when adopting the portal element into other top-level browsing contexts; developers are required to change and then set back the `src` content attribute to re-trigger a load in such cases.

jeremyroman:
Hopefully this explains how the adopting steps prevent you from embedding a portal inside an <iframe>.

Resolves #82.

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/portals/pull/103.html" title="Last updated on Apr 25, 2019, 3:21 PM UTC (ca61ba4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/103/9766d82...jeremyroman:ca61ba4.html" title="Last updated on Apr 25, 2019, 3:21 PM UTC (ca61ba4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -479,9 +479,15 @@ spec: fetch; urlPrefix: https://fetch.spec.whatwg.org/
   1. [=close a portal element|Close=] |element|.
 
   <div class="note">
-    Since <{portal}> elements can have only a [=guest browsing context=] while
-    their [=node document|document=] belongs to a [=top-level browsing
-    context=], it is impossible to embed a [=portal browsing context=] in a
+    In particular, this means a <{portal}> element loses its [=guest browsing
+    context=] if it is moved to the [=active document=] of a [=nested browsing
+    context=].
+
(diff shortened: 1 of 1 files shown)


jeremyroman:
@domenic

PR #105, opened 2019-04-29
Adding demos directory && portal-embed-demo

uskay:
This is a PR for adding demos to portals repository.

- The directory structure will be: `root/demos/{demo_project_name}`
- portal-embed-demo demonstrates how Portals can enable a seamless user experience between a website and a third party embed content. 
- Adding detailed explainer to README.md so that developers can easily play around with it

CC: @jeremyroman @lucasgadani @KenjiBaheux


demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,148 @@
...
+const portal = document.createElement('portal');
+portal.src = 'https://example.com';
+```
+> Demo code reference: [creating a Portal element](https://github.com/WICG/portals/blob/master/demos/portal-embed-demo/public/js/portalog/portals-controller.js#L37) 
+
+For the Portal content side, you can judge if the page is used inside a Portal and customize the UI.
+```javascript
+// Check if window.portalHost is available
+if(window.portalHost){
+   // Customize the UI when being
```
jeremyroman:
```suggestion
   // Customize the UI when being embedded as a portal.
```

jeremyroman: Maybe "Customize the UI when being embedded as a portal." or similar?

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,148 @@
...
+   // embedded as portals
+}
+```
+> Demo code reference: [Check if window.portalHost is avaiable](https://github.com/WICG/portals/blob/master/demos/portal-embed-demo/public/js/ttt/portals-controller.js#L198) and [change the style](https://github.com/WICG/portals/blob/master/demos/portal-embed-demo/public/js/ttt/portals-controller.js#L200)
+
+For now, Portal will not process user input. If you want to interact with the Portal (like playing the audio in the demo), use `postMessage`/`onmessage`.
```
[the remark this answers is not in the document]
jeremyroman: also, nit: it seems a little excessive to explicitly point out that `onmessage` is used to handle incoming message. IMHO scans better if you just say "use `postMessage`", but not sure.

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,148 @@
...
+```javascript
+// Send message from the portal element
+const portal = document.querySelector('portal');
+portal.postMessage({someKey: someValue}, ORIGIN);
+
+// Receive message via window.portalHost
+window.portalHost.addEventListener('message', evt => {
+   const data = evt.data;
+   // do the actual following
```
jeremyroman: This comment makes sense in the original context, but seems a little out of place here.

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,148 @@
...
+// Listen to the portalactivate event
+window.addEventListener('portalactivate', evt => {
+   // ... and creatively use the predecessor
+   const portal = evt.adoptPredecessor();
+   document.querySelector('someElm').appendChild(portal);
+});
+```
+> Demo code reference: [listening to `portalactivate`](https://github.com/WICG/portals/blob/master/demos/portal-embed-demo/public/js/ttt/portals-controller.js#L144) and [reusing the predecessor](https://github.com/WICG/portals/blob/master/demos/portal-embed-demo/public/js/ttt/portals-controller.js#L152)
+
+On the predcessor side, `activate` will resolve with an `undefined` `Promise` when the Portal activation has completed. If it was adopted as a predecessor, you can start using `window.portalHost`.
```
jeremyroman:
```suggestion
On the predecessor side, `activate` will resolve with an `undefined` `Promise` when the Portal activation has completed. If it was adopted as a predecessor, you can start using `window.portalHost`.
```

Up to you, but I'm also not sure it's interesting to point out that a `Promise<void>` resolves with `undefined`. Maybe just say "`activate` returns a promise that resolves when activation has completed."?

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,148 @@
+# portal-embed-demo
+This demo demonstrates how Portals can enable a seamless user experience between a website and a third party embed content. Be creative and have fun!!
+<br/>
```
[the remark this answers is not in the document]
uskay: I wanted extra line break break before header lines but I agree that it's making it messy. Fixed it.
https://github.com/WICG/portals/pull/105/commits/192e74c821a6f2de6dfbfc61d591a3da2035d55c#diff-75118edbb0db728bc0648cb6e01c9c2bL3

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,148 @@
...
+```bash
+$ npm run demo
+📝 PORTALOG has launched: http://localhost:3000?portalport=3001
+🎧 TTT Archive has launched: http://localhost:3001
+```
+
+### 2. Open a browser that supports Portals
+As of May 2019, Chrome Canary is the only platform that supports Portals. You can try out Portals in Chrome Canary by flipping an experimental flag (chrome://flags/#enable-portals).
+
+### 3. Access [http://localhost:3000/](http://localhost:3000/)
```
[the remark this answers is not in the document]
uskay: I wanted to make it easy to change the port of the two local servers. 
- `server.js ` has a default port settings but developers might want to change the port being used: https://github.com/WICG/portals/pull/105/commits/af7a1e8b2d8bd7f64b42ebc1515cd6cfb06f8b68#diff-ffc49d078375ee2224fd41a7bbb1d456R3
- and `public/js/portalog/portals-controller.js` checks if it's using other than the default 3001 for TTT Archive and replaces the URL to use for portals: https://github.com/WICG/portals/pull/105/commits/af7a1e8b2d8bd7f64b42ebc1515cd6cfb06f8b68#diff-aff8d943c491a10805f3c6c4b735b032R201 

Since in this application, both `server.js` and portalog needs to know the port used for TTT Archive, the only thing I could think of to easily avoid hard coding was to pass the port with a query parameter.

Does this make sense?

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,148 @@
...
+<!-- You can use it like iframes -->
+<portal src='https://example.com'></portal>
+```
+OR
+```javascript
+// From JavaScript
+const portal = document.createElement('portal');
+portal.src = 'https://example.com';
+```
+> Demo code reference: [creating a Portal element](https://github.com/WICG/portals/blob/master/demos/portal-embed-demo/public/js/portalog/portals-controller.js#L37)
```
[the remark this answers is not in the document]
uskay: Thank you so much for this! Changed all the links to relative path.
https://github.com/WICG/portals/pull/105/commits/192e74c821a6f2de6dfbfc61d591a3da2035d55c#diff-75118edbb0db728bc0648cb6e01c9c2bR71

demos/portal-embed-demo/README.md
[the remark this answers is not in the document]
uskay: I've changed
- s/Portal/portal/g ... changing it to lowercase e.g. https://github.com/WICG/portals/pull/105/commits/192e74c821a6f2de6dfbfc61d591a3da2035d55c#diff-75118edbb0db728bc0648cb6e01c9c2bR73
- Adding an article to each one of them e.g. a/the, or change it to "portals" e.g. https://github.com/WICG/portals/pull/105/commits/192e74c821a6f2de6dfbfc61d591a3da2035d55c#diff-75118edbb0db728bc0648cb6e01c9c2bR82

demos/portal-embed-demo/README.md
uskay: Committed your suggestion.
https://github.com/WICG/portals/pull/105/commits/192e74c821a6f2de6dfbfc61d591a3da2035d55c#diff-75118edbb0db728bc0648cb6e01c9c2bR77

demos/portal-embed-demo/README.md
uskay: Thank you so much! Yes, this is not good. Fixed it to "handle the event"
https://github.com/WICG/portals/pull/105/commits/192e74c821a6f2de6dfbfc61d591a3da2035d55c#diff-75118edbb0db728bc0648cb6e01c9c2bR91

uskay:
> Mostly reviewed the README. Wasn't clear to me how closely the demo code ought to be reviewed (it's not production code so I'm not sure what the norm is here).

I understand this is a huge ask but I would love to have a review on the code (although it's not a production code). I do have other demos I would love to contribute in the future, and I would want the quality to hit your expectations.

demos/portal-embed-demo/README.md
uskay: Thank you for the comment! I wasn't so used to using void Promises in other cases so was adding this comment but as you mentioned since it's <void> it's obvious it will not return something. Changed the comments.
https://github.com/WICG/portals/pull/105/commits/cfabda684eec194839e14ed66a384d869d641852

uskay:
> Also not obvious where the images come from (license?).

Thank you for the comment on this. Added credit in README.md.
https://github.com/WICG/portals/pull/105/commits/8743cee98483dee998342a74a35aaba17a64fa6f

demos/portal-embed-demo/README.md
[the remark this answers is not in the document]
jeremyroman: Sorry for not being clear. I'm wondering if the instructions should point to http://localhost:3000/?portalport=3001, to be consistent with what the npm command spits out, or  alternatively, instead instruct readers to click the link npm emits.

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+
+        // clean up the root
+        this._remove(this.root.querySelector('portal'));
+
+        if (predecessor) {
+            // If the optional predecessor parameter was set
+            // add the predecessor to the root
+            this.portal = predecessor;
+        } else {
+            // Or else, create new portal
```
jeremyroman: Having this handle two paths, even though they both only have one call site, seems a little confusing. I would suggest that instead, we structure the code as two separate cases. All that we need to do is replace the old portal with the new portal, hook up any event listeners (or we can avoid that by just registering the event listener on the parent node thanks to propagation), and restore the player UI.

```javascript
/**
 * Creates a new <portal> element and player UI.
 */
populateEmbedContainer() {
    this.portal = document.createElement('portal');
    this.portal.src = this.src;
    this.playerUI = this._genPortalMessageUI();
    this.embedContainer.append(this.portal, this.playerUI);

    // When the portal is clicked, start animating it.
    this.embedContainer.addEventListener('click', evt => {
        if (evt.target === this.portal)
            this.animateAndActivate();
    });
}

/**
 * Animates the embedded content to smoothly activate it.
 */
animateAndActivate() {
    this.playerUI.style.display = 'none';
    // (current logic to start the CSS transition)
}

/**
 * Reinstalls the embedded portal after returning to PORTALOG.
 * @param {HTMLPortalElement} predecessor A portal element containing the embedded page.
 */
returnFromEmbed(predecessor) {
    this.playerUI.style.display = '';
    this.portal.replaceWith(predecessor);
    this.portal = predecessor;
}
```

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+            this.root.style.width = '95%';
+            this.root.style.paddingTop = 'calc(95% * 0.65)';
+            this.root.style.left = 'calc((100% - 95%)/2)';
+            this.portal.postMessage({ control: 'hide' }, this.origin);
+        });
+
+    }
+
+    /**
+     * Adding event listeners
```
jeremyroman: This description is kind of vague. It might be worth just inlining this into the function that creates the DOM (if you take my suggestion above and split the adoptPredecessor case out to just `replaceWith`. At that point, actually, maybe both just ought to be in the `PortalsController` constructor, since you always want to call it after constructing that class.

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+                        name: 'Yusuke Utsunomiya',
+                        photoSrc: '/img/profile.png',
+                        initialY: this.initialY,
+                        activatedWidth: this.root.getBoundingClientRect().width,
+                        initialWidth: this.initialWidth,
+                    }
+                }).then((_) => {
+                    // Resolves with undefined when adoped as a predecessor
+                    // Check if window.portalHost is present (just in case)
+                    if(!window.portalHost){
```
jeremyroman:
```suggestion
                    if (!window.portalHost) {
```

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+                    window.portalHost.addEventListener('message', (evt) => {
+                        const isFollowed = evt.data.isFollowed;
+                        this._changeFollowStatus(!isFollowed);
+                    });
+                });
+
+                this._resetRoot();
+            }
+        });
+        document.querySelector('#follow').addEventListener('click', (evt) => {
```
jeremyroman: If we have time, these bits (the follow button and player UI) probably ought to be split out from the main `PortalsController` class; it does too many things that aren't super directly related.

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+                    // Check if window.portalHost is present (just in case)
+                    if(!window.portalHost){
+                        return;
+                    }
+
+                    // hide the scroll bar
+                    document.body.style.overflow = 'hidden';
+
+                    // Listen to messages (follow/unfollow)
+                    window.portalHost.addEventListener('message', (evt) => {
```
jeremyroman: This listener is added every time we swap, so we end up with several copies of this listener. We should either remove it every time we are activated, or track that we have already added it and not do so again.

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+                        const isFollowed = evt.data.isFollowed;
+                        this._changeFollowStatus(!isFollowed);
+                    });
+                });
+
+                this._resetRoot();
```
jeremyroman: It's not obvious what it means to "reset the root". It looks like this resets the animated properties, so rename and update function comment accordingly.

demos/portal-embed-demo/public/js/ttt/ttt-reco.js
```diff
@@ -0,0 +1,169 @@
...
+                </div>
+            </div>
+        `;
+    }
+
+    /**
+     * Adding event listeners for the buttons
+     */
+    _hookEvents() {
+        [...this.root.querySelectorAll('.add-button')].map(elm => {
```
jeremyroman: We don't use the result of this map, so it would be more idiomatic to just write:

```javascript
for (let elm of this.root.querySelectorAll('.add-button')) {
    /* ... */
}
```

demos/portal-embed-demo/server.js
```diff
@@ -0,0 +1,20 @@
+const express = require('express');
+const PORTALOG_PORT = 3000;
+const TTT_ARCHIVE_PORT = 3001;
+
+const genServer = (serverName, port, rootHTML) => {
```
jeremyroman: consistent style with client JS? (e.g. 4sp or 2sp in both places)

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,145 @@
...
+> Demo code reference: [listening to `portalactivate`](public/js/ttt/portals-controller.js#L144) and [reusing the predecessor](public/js/ttt/portals-controller.js#L152)
+
+`activate` returns a promise that resolves when activation has completed.
+```javascript
+// The activate function returns a Promise.
+// When the promise resolves, it means that the portal has been activated.
+// If this document was adopted by it, then window.portalHost will exist
+// When the promise resolves, it means the page was adopted as a predecessor
+portal.activate().then(_ => {
+   // check if portalHost is available
```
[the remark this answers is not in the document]
uskay: That is a great point, thank you!

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,145 @@
...
+   if(window.portalHost){
```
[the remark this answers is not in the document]
uskay: Sorry, for this. Will check others as well.

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,145 @@
...
+// From JavaScript
+const portal = document.createElement('portal');
+portal.src = 'https://example.com';
+```
+> Demo code reference: [creating a portal element](public/js/portalog/portals-controller.js#L37) 
+
+For the portal content side, you can judge if the page is used inside a portal and customize the UI.
+```javascript
+// Check if window.portalHost is available
+if(window.portalHost){
```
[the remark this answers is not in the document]
uskay: Thank you so much for this as well and apologies. I will check other if statement as well.

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+    constructor(root, origin, path, anmiateY) {
+        this.root = root;
+        this.origin = origin;
+        this.path = path;
+        this.anmiateY = anmiateY;
+    }
+
+    /**
+     * Embedding portal element to the root.
+     * @param {HTMLPortalElement} predecessor - optional
```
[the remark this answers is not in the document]
uskay: I didn't know this. I'm learning a lot. Thank you!

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+ */
+class PortalsController {
+
+    /**
+     * Initiating paratmeter
+     * @constructor
+     * @param {HTMLElement} root - the root element
+     * @param {String} origin - origin of portal
+     * @param {String} path - path of portal
+     * @param {Number} anmiateY - A target position to animate in px
```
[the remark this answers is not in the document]
uskay: Thank you for this! Understood and after all I've changed to targetY!

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+        window.addEventListener('portalactivate', (evt) => {
+            // reset the overflow style to default
+            document.body.style.overflow = '';
+            // embed predecessor
+            this.embed(evt.adoptPredecessor());
+        });
+
+        // Event fires after the portal on-click animation finishes
+        this.root.addEventListener('transitionend', (evt) => {
+            if (evt.propertyName === 'top') {
```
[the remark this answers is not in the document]
uskay: Makes sense! Do you think keeping the bracket is good even if you can skip writing it in this case?
```javascript
if (evt.propertyName !== 'top') {
    return;
}
```

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+                    data: {
+                        followed: isFollowed,
+                        name: 'Yusuke Utsunomiya',
+                        photoSrc: '/img/profile.png',
+                        initialY: this.initialY,
+                        activatedWidth: this.root.getBoundingClientRect().width,
+                        initialWidth: this.initialWidth,
+                    }
+                }).then((_) => {
+                    // Resolves with undefined when adoped as a predecessor
```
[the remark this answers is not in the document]
uskay: SGTM!

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+        controller.appendChild(next);
+
+        return controller;
+    }
+
+    /**
+     * Removing an element
+     * @param {HTMLElement} elm 
+     */
+    _remove(elm) {
```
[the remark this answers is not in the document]
uskay: Yes, I will remove it.

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+    const matchResult = location.search.match(/portalport=(\d{4})/);
+    if(matchResult[1]){
+        portalPort = matchResult[1];
+    }
+}
+
+if('HTMLPortalElement' in window){
+    // Create instance
+    const portalsController = new PortalsController(
+        document.querySelector('#embed'),
```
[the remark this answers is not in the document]
uskay: Yes, I will do that:

```javascript
let embedContainer = document.querySelector('#embed');
if ('HTMLPortalElement' in window) {
    // Create instance
    const portalsController = new PortalsController(
        embedContainer,
        `http://localhost:${portalPort}`,
        170
    );
...
    embedContainer.appendChild(iframe, link);
```

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+        `http://localhost:${portalPort}`,
+        '/',
```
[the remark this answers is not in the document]
uskay: Agreed, I'm removing the path parameter and consolidating to the URL. Will use the URL interface.

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+        170
```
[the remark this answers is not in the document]
uskay: Agreed. Since it's only used in the animation function, I'll replace it like this
``` javascript
    /**
     * Animates the embedded content to smoothly activate it.
     */
    animateAndActivate() {
        // A Y poistion that needs to be animated to
        // TODO: retrieve this value from the portal via postMessage
        // in the future
        const TARGET_Y = 170;
```
I kind of wanted to loosely couple this value from the PortalController class itself since the position requirement is from the TTT Archive side and not PORTALOG. So, added a TODO comment that in the future, it would be good to retrieve this value from messages. WDTY?

demos/portal-embed-demo/public/js/portalog/portals-controller.js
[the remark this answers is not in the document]
jeremyroman: I'm fine with either. (I was in that habit because it is what we do in Chromium C++.)

Now that I look, [Google JS style](https://google.github.io/styleguide/jsguide.html#formatting-braces) says that short conditions can be put on one line, or braces can be used, but not broken as I showed. So instead, either of these would probably be ideal:

```javascript
if (evt.propertyName !== 'top') return;
```
or
```javascript
if (evt.propertyName !== 'top') {
    return;
}
```

(unfortunately since I don't write "real" JS often, I'm a little rusty on recommended style in some places)

demos/portal-embed-demo/public/js/portalog/portals-controller.js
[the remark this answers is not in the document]
jeremyroman: Might not be a bad idea, though it might be overkill for demo purposes. Happy to defer to your judgment.

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+     * @param {String} id - audio ID
+     */
+    _createNewAudio(id) {
+
+        // Simply replace the exisiting audio element with a new one
+        if(this.root.querySelector('audio')){
+            this.root.removeChild(this.root.querySelector('audio'));
+        }
+        const audio = document.createElement('audio');
+        const source = document.createElement('source');
```
jeremyroman: Since we only ever have one source, it's simpler to just set `src` on the media element. (The HTML spec also recommends this approach as you can change what is playing on a media element by changing its `src`, but not by modifying its `<source>` children: https://html.spec.whatwg.org/#the-source-element)

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+    _hasPrev() {
+        return this.playlistIndex !== 0;
+    }
+
+    /**
+     * Checks if the controller has a next audio to play
+     * @returns {boolean}
+     */
+    _hasNext() {
+        return this.playlist.length > 1
```
jeremyroman: Is the first part of this condition actually necessary given the second part? ISTM that this should suffice:

```javascript
return this.playlistIndex < this.playlist.length - 1;
```

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+            = `${audioList[id].title} - Totally Tooling Tips`;
+        this.axis.style.transition = '';
+        this.axis.style.left = '0px';
+    }
+
+    /**
+     * Update the prev/next button status
+     */
+    _updateController() {
+        if(this._hasPrev()) {
```
jeremyroman:
```suggestion
        if (this._hasPrev()) {
```

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+            this.prevButton.style.opacity = 1.0;
+        } else {
+            this.prevButton.style.opacity = 0.3;
+        }
+        if(this._hasNext()){
```
jeremyroman:
```suggestion
        if (this._hasNext()){
```

demos/portal-embed-demo/public/js/ttt/audio-controller.js
jeremyroman: nit: Also, I wonder if we couldn't just define a "disabled" class in the CSS and make this implementation shorter:

```javascript
this.prevButton.classList.toggle('disabled', !this._hasPrev());
this.prevButton.classList.toggle('disabled', !this._hasNext());
```

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+     */
+    get template() {
+        return `
+            <div class='title'></div>
+            <div class='timeline'>
+                <div class='axis'></div>
+            </div>
+            <div class='controllers'>
+                <div class='controller'>
+                    <img class='prev flip-horizontal' src='/img/next.png'>
```
jeremyroman: I assume for the purposes of this demo we're not going to include proper a11y? It's a little unfortunate but maybe makes sense given time constraints.

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+            this.axis.style.left = `${left}px`;
+            this.isPlaying = true;
+        }
+        // https://developers.google.com/web/updates/2017/06/play-request-was-interrupted
+        const promise = this.audio.play();
+        if(promise !== undefined) {
+            promise.then(_ => {
+                startAxis();
+            }).catch(e => {
+                console.log(e);
```
jeremyroman: Aren't unhandled promise rejections logged to the console regardless?

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+        const startAxis = _ => {
+            // Start the progress bar
+            const left = this.timeline.offsetWidth;
+            this.axis.style.transition = `${this.durationSec}s linear`;
+            this.axis.style.left = `${left}px`;
+            this.isPlaying = true;
+        }
+        // https://developers.google.com/web/updates/2017/06/play-request-was-interrupted
+        const promise = this.audio.play();
+        if(promise !== undefined) {
```
jeremyroman: Since it seems we want to treat the no-promise and promise-resolved case equivalently, we can equivalently just write:

```javascript
Promise.resolve(this.audio.play()).then(_ => {
    // Start the progress bar.
});
```

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    /**
+     * HTML template
+     */
+    get template() {
+        return `
+            <div class='title'></div>
+            <div class='timeline'>
+                <div class='axis'></div>
```
jeremyroman: "axis" is a little confusing as a name. Why not call it "progress bar" or similar?

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+            }
+        `;
+    }
+
+    /**
+     * HTML template
+     */
+    get template() {
+        return `
+            <div class='title'></div>
```
jeremyroman: nit: double-quotes are usually used for HTML attributes (and you're inside backticks, so you should be able to use them without extra escaping)

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+            return;
+        }
+        startAxis();
+    }
+
+    /**
+     * Controlling pause button
+     */
+    pause() {
+        if(!this._isPlaying()) {
```
jeremyroman:
```suggestion
        if (!this._isPlaying()) {
```

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+
+    /**
+     * Controlling play button
+     */
+    play() {
+        this.playButton.src = '/img/pause.png';
+        const startAxis = _ => {
+            // Start the progress bar
+            const left = this.timeline.offsetWidth;
+            this.axis.style.transition = `${this.durationSec}s linear`;
```
jeremyroman: I don't think this timing is correct if we've paused the audio half-way through the file. The progress bar will progress such that it would reach the end after `this.durationSec` seconds, but only `this.durationSec / 2` actually remain at that point. This should probably be set based on the actual seek position.

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+    /**
+     * Remove audio from playlist
+     * @param {String} id - audio ID
+     */
+    removePlaylist(id) {
+        // Simply create new playlist filtering the ID to remove
+        const newPlaylist = [];
+        this.playlist
+            .filter(audioId => audioId !== id)
+            .map(audioId => newPlaylist.push(audioId));
```
jeremyroman: `filter` already produces the array you want:

```javascript
this.playlist = this.playlist.filter(audioId => audioId !== id);
```
`splice` would be more efficient, yet, assuming the ID can only occur once, but maybe that's not crucial for this demo:

```javascript
const idx = this.playlist.indexOf(audioId);
if (idx !== -1) this.playlist.splice(idx, 1);
```

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+            this.play();
+        }
+        this._updateController();
+    }
+
+    /**
+     * Add audio to playlist
+     * @param {String} id - audio ID
+     */
+    addPlaylist(id) {
```
jeremyroman: Name is confusing; sounds like it adds a playlist. Maybe `addToPlaylist` or `addTrack`?

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+        this.playlist.push(id);
+        this._updateController();
+    }
+
+    /**
+     * Remove audio from playlist
+     * @param {String} id - audio ID
+     */
+    removePlaylist(id) {
```
jeremyroman: similar

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+    show() {
+        this.style.transition = 'bottom 0.2s';
+        this.style.bottom = '0px';
+    }
+
+    /**
+     * Adding anything required to do on portal activate
+     */
+    handleActivation() {
+        if(this._isPlaying()) {
```
jeremyroman:
```suggestion
        if (this._isPlaying()) {
```

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    /**
+     * Embedding portal element to the root.
+     * @param {HTMLPortalElement} predecessor - optional
+     */
+    embed(predecessor) {
+
+        // clean up the root
+        this._remove(this.root.querySelector('portal'));
```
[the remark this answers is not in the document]
uskay: Read thru other comments. I agree with just using `this.portal.replaceWith(predecessor)`. Thank you!

demos/portal-embed-demo/public/js/portalog/portals-controller.js
uskay: Agreed. This makes things a lot easier to read! Will keep things minimum with scoping the concerns of each function.

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+        }
+        const messageUI = this._genPortalMessageUI();
+        this.root.appendChild(this.portal);
+        this.root.appendChild(messageUI);
+
+        // Add animation on-click
+        this.portal.addEventListener('click', (evt) => {
+            this.initialY = this.root.getBoundingClientRect().y;
+            this.initialWidth = this.root.getBoundingClientRect().width;
+            this._remove(this.root.querySelector('.message-controller'));
```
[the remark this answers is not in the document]
uskay: The initial design was, I kind of wanted to not care about the state of the PORTALOG when going back and forth with TTT Archive. So the concept was to rebuild the whole stuff when coming back. But as you pointed out in other comments, I agree it's better and easier to read with reusing the existing elements. I'll go with `display: none`.

demos/portal-embed-demo/public/js/ttt/writer-follow.js
```diff
@@ -0,0 +1,118 @@
...
+     * Adding event listeners for the buttons
+     */
+    _hookEvents() {
+        const follow = this.root.querySelector('.follow');
+        const ORIGIN = 'http://localhost:3000';
+        // On-click event for the follow button
+        follow.addEventListener('click', evt => {
+            const isFollowed = follow.classList.contains('followed');
+            const portal = document.querySelector('portal');
+            if(isFollowed){
```
jeremyroman:
```suggestion
            if (isFollowed) {
```

demos/portal-embed-demo/public/js/ttt/writer-follow.js
```diff
@@ -0,0 +1,118 @@
...
+            </div>
+        `
+    }
+
+    /**
+     * Adding event listeners for the buttons
+     */
+    _hookEvents() {
+        const follow = this.root.querySelector('.follow');
+        const ORIGIN = 'http://localhost:3000';
```
jeremyroman: While constant in this scope, it's not statically constant, so suggest lowercase `origin`.

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+
+    /**
+     * Adding event listeners
+     */
+    hookEvents() {
+        
+        // Event fires when coming back from TTT Archive
+        window.addEventListener('portalactivate', (evt) => {
+            // reset the overflow style to default
+            document.body.style.overflow = '';
```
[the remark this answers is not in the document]
uskay: Agreed. Changed it to:

```javascript
// hide the scroll bar so that it won't show when used as a predecessor 
document.body.classList.add('hide-scroll-bars');
```

```javascript
// show the scroll bar when coming back
document.body.classList.remove('hide-scroll-bars');
```

demos/portal-embed-demo/public/js/portalog/portals-controller.js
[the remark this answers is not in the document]
uskay: I will look into the style guide. I'm used to keeping a bracket and that's implemented elsewhere so will implement the latter one!

demos/portal-embed-demo/public/js/portalog/portals-controller.js
uskay: Changed it to:
```javascript
// Reset the position of the container after the portal activates
this._resetPositionOfEmbedContainer();
```

demos/portal-embed-demo/public/js/portalog/portals-controller.js
uskay: Agreed. For the interest of time, I added a TODO in the comment. I think I'll put it in a backlog and have those refactoring work separately after this week.

```javascript
// TODO: Separate the follow operations from the PortalsController
// https://github.com/WICG/portals/pull/105/files#r279998963
document.querySelector('#follow').addEventListener('click', (evt) => {
```

demos/portal-embed-demo/public/js/ttt/portals-controller.js
```diff
@@ -0,0 +1,214 @@
...
+        this.audioController = document.querySelector('audio-controller');
+
+        // Add event listeners
+        this._hookEvents();
+    }
+
+    /**
+     * Controll all the UI/animations with given display mode
+     * @param {String} mode 
+     * @param {Object} option
```
jeremyroman: If you split the method, you can get proper typing for the arguments that apply to case.

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,239 @@
...
+            follow.classList.add('followed');
+            follow.textContent = 'Following';
+        }
+    };
+
+}
+
+let portalPort = 3001;
+if(location.search) {
+    const matchResult = location.search.match(/portalport=(\d{4})/);
```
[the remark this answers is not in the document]
uskay: That is much cleaner!! Thank you!! I added the constant to the top of the file:

```javascript
/**
 * Default port used for the portal
 */
const DEFAULT_PORTAL_PORT = 3001;
```

demos/portal-embed-demo/public/js/portalog/portals-controller.js
uskay: Agreed. I moved it to the constructor. (I don't remember why I designed it as a public function... so probably unnecessary stuff)

demos/portal-embed-demo/public/js/portalog/portals-controller.js
uskay: That is a great point!! In my understanding, there is no good way to check if the listener was added to an element (other than adding additional data structure to manage that state), I simply added something like below.

In the constructor...
```javascript
    constructor(embedContainer, src) {
        this.embedContainer = embedContainer;
        this.src = src;
        this.isPortalHostListenerAdded = false;
```
And after activating...
```
// don't add event listeners if it was already added
if(this.isPortalHostListenerAdded) {
    return;
}

// Listen to messages (follow/unfollow)
window.portalHost.addEventListener('message', (evt) => {
    const isFollowed = evt.data.isFollowed;
    this._changeFollowStatus(!isFollowed);
});
this.isPortalHostListenerAdded = true;
```

Do you think this is okay? I kind of feel like there could be more smart way to do this than using dedicated flag but as this is a demo, should be sufficient IMO.

demos/portal-embed-demo/public/js/portalog/portals-controller.js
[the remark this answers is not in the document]
uskay: I'll keep the TODO and work on it next week as a refactoring task (and check then if that's necessary).

demos/portal-embed-demo/public/js/portalog/portals-controller.js
jeremyroman: sgtm

demos/portal-embed-demo/README.md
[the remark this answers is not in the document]
uskay: Got you point. Changed it to access ` http://localhost:3000/?portalport=3001`.

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,145 @@
...
+```javascript
+// From JavaScript
+const portal = document.createElement('portal');
+portal.src = 'https://example.com';
+```
+> Demo code reference: [creating a portal element](public/js/portalog/portals-controller.js#L37) 
+
+For the portal content side, you can judge if the page is used inside a portal and customize the UI.
+```javascript
+// Check if window.portalHost is available
```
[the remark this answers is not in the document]
uskay: That is so true. I'll just delete this comment.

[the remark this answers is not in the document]
uskay: Ahh I think it's nice to have your suggestion: `// Detect whether this page is hosted in a portal`. Replaced with this.

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,145 @@
...
+if(window.portalHost){
+   // Customize the UI when being embedded as a portal
```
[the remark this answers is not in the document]
uskay: Thank you for the comment! Changed all the snippets to 2 space indent!

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,145 @@
...
+}
+```
+> Demo code reference: [Check if window.portalHost is available](public/js/ttt/portals-controller.js#L198) and [change the style](public/js/ttt/portals-controller.js#L200)
```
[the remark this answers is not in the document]
uskay: Yes 😂I think it's useful too so will keep track of it for the initial commit. Might develop some tool to do this.

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,145 @@
...
+window.portalHost.addEventListener('message', evt => {
+   const data = evt.data.someKey;
+   // handle the event
+});
+```
+> Demo code reference: interacting with the audio player ([sending messages](public/js/portalog/portals-controller.js#L135) and [receiving messages](public/js/ttt/portals-controller.js#L183))
+
+When the user decides to navigate the the portal content i.e. click, it is a good opportunity to animate the portal and then call the `activate` function. User will be navigated to the portal content seamlessly (but the URL changes). You can also see that the session of the portal content is consistent and the audio keeps playing even after this activation.
+```javascript
+// do some fancy animations and..
```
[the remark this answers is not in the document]
uskay: Agreed! Changed it to:
```javascript
// do some fancy animations and after animations are complete, activate the portal.
const portal = document.querySelector('portal');
```

demos/portal-embed-demo/README.md
```diff
@@ -0,0 +1,145 @@
...
+`activate` returns a promise that resolves when activation has completed.
+```javascript
+// The activate function returns a Promise.
+// When the promise resolves, it means that the portal has been activated.
+// If this document was adopted by it, then window.portalHost will exist
+// When the promise resolves, it means the page was adopted as a predecessor
+portal.activate().then(_ => {
+   // check if portalHost is available
+   if(window.portalHost){
+        // The page was adopted as a predecessor. You can start
```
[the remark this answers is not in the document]
uskay: Fixed it!

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: Thank you for this! I changed all (including ttt-reco, writer-follow) the template() HTML using double quotes!

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: Thank you! I didn't know what to express that small progress bar's dot was just calling axis. Will change it to progress-bar.

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: let me work on a11y as a refactoring work next week.

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: ...!! This is much more readable!!

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: Yes, as this is for demo use, I'll remove this catch.

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: Changed all the code styles in the file (will check other files as well).

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: Changed it to: `addToPlaylist`

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: Changed it to `removeFromPlaylist`

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: Ahhhh, please let me use your recommendation.
``` javascript
        // Remove given id from playlist
        const index = this.playlist.indexOf(audioId);
        if (index !== -1) this.playlist.splice(index, 1);
```

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: It actually stops when pausing.
```javascript
pause() {
  ...
  // Stop the progress bar
  this.progressBar.style.left = `${this._getProgressBarCurrentLeft()}px`;
}
```
I'm wondering how I can implement using the actual seek position. I wasn't sure if we should use setInterval for this (wondering the performance cost of scheduling running a script per second). If there was some observables I can use, that could be another option but could you kindly advice me here?
```javascript
const updateProgressBar = _ => {
  this.progressBar.style.left = this.timeline.offsetWidth * this.audio.currentTime/this.audio.duration;
}
setInterval(updateProgressBar, 1000);
``

File: demos/portal-embed-demo/public/js/ttt/audio-controller.js [hunk 54]
```diff
@@ -0,0 +1,369 @@
...
+        this.playButton.addEventListener('click', evt => {
+            if(this._isPlaying()){
+                this.pause();
+                return;
+            }
+            this.play();
+        });
+        this.nextButton.addEventListener('click', evt => this.next());
+        this.prevButton.addEventListener('click', evt => this.prev());
+        this.audio.addEventListener("ended", evt => {
```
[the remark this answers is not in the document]
uskay: ...! Thanks for this. Fixed the bug.

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: Understood. Thank you for this comment. Fixed it.
```javascript
audio.src = `/mp3/${id}.mp3`;
this.root.appendChild(audio);
```

demos/portal-embed-demo/public/js/ttt/audio-controller.js
jeremyroman: > It actually stops when pausing.
> 
> ```js
> pause() {
>   ...
>   // Stop the progress bar
>   this.progressBar.style.left = `${this._getProgressBarCurrentLeft()}px`;
> }
> ```

Right, but what I mean is (even without accounting for audio buffering and other possible sources of delay): suppose the audio clip is 30 seconds long.1
1. Click play. 30-second transition starts (`this.durationSec`).
2. Wait 15 seconds.
3. Click pause. Transition pauses; style is kept at 50% of the progress bar width.
4. Click play. 30-second transition starts (`this.durationSec`).
5. Wait 15 seconds.
6. Audio clip finishes, but progress bar is only 75% to the right (50% of the transition from 50% to 100% has completed).

> I'm wondering how I can implement using the actual seek position. I wasn't sure if we should use setInterval for this (wondering the performance cost of scheduling running a script per second). If there was some observables I can use, that could be another option but could you kindly advice me here?
> 
> ```js
> const updateProgressBar = _ => {
>   this.progressBar.style.left = this.timeline.offsetWidth * this.audio.currentTime/this.audio.duration;
> }
> setInterval(updateProgressBar, 1000);
> ``
> ```

I believe this is what the [`timeupdate` event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/timeupdate_event) is for (though I've never used it).

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: True. Unnecessary. Fixed it to your suggestion.

demos/portal-embed-demo/public/js/ttt/audio-controller.js
```diff
@@ -0,0 +1,369 @@
...
+        const offset = (playerWidth - timelineWidth)/2;
+        const rectLeft = this.axis.getBoundingClientRect().left;
+        return rectLeft - offset;
+    }
+
+    /**
+     * Checks if the controller is currently playing an audio
+     * @returns {boolean}
+     */
+    _isPlaying() {
```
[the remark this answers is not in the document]
uskay: Yes. Initially I was managing the state with a custom attribute set to the play button like `this.button.setAttribute("isPlaying", true/false)` and wanted to have that operation/checking in a separate function. But since I ended up using the class member, I'll just remove this function.

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: This works beautifully. I changed it to your recommendation and 1) removed the initial style of the button (which was set to opacity 0.3) 2) called _updateController() in the constructor and 3) removed any unnecessary operations around button opacity.

jeremyroman: (just realized I meant to write `nextButton` on the second line, but I assume you caught that mistake :wink:)

demos/portal-embed-demo/public/js/ttt/audio-controller.js
uskay: THAT IS SO TRUE. I ended up implementing something like this: 
```javascript
        const startProgressBar = _ => {
            // Start the progress bar
            this.isPlaying = true;
            this.audio.addEventListener('timeupdate', evt => {
                this.progressBar.style.left = 
                    `${this.timeline.offsetWidth * this.audio.currentTime/this.audio.duration}px`;
            });
        }
```
And I get to delete `handleActivation` and ` _getProgressBarCurrentLeft` and remove all the unnecessary complexity!!!!!

Thank you!🙇🙇🙇

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,236 @@
...
+
+        // Event fires after the portal on-click animation finishes
+        this.embedContainer.addEventListener('transitionend', (evt) => {
+
+            // We wait until the top transition finishes
+            if (evt.propertyName !== 'top') {
+                return;
+            }
+
+            const isFollowed = document.querySelector('#follow')
```
jeremyroman: nit: suggest moving the rest of this function into a function on PortalController (maybe `activateAfterAnimation` or something like that), probably placed after `activateAndAnimate`, to both make it clear that it's what follows that animation and to make this constructor a little more focused on hooking up the listeners.

Then the members end up following the usual order of events:
* construct
* populate embed container
* animate
* actually activate

demos/portal-embed-demo/public/js/portalog/portals-controller.js
```diff
@@ -0,0 +1,236 @@
...
+                data: {
+                    followed: isFollowed,
+                    name: 'Yusuke Utsunomiya',
+                    photoSrc: '/img/profile.png',
+                    initialY: this.initialY,
+                    activatedWidth: this.embedContainer.getBoundingClientRect().width,
+                    initialWidth: this.initialWidth,
+                }
+            }).then((_) => {
+                // Check if window.portalHost is present (just in case)
```
jeremyroman:
```suggestion
                // Check if this page was adopted by the embedded content.
```

demos/portal-embed-demo/public/js/ttt/audio-controller.js
jeremyroman: Nice!

We can probably attach the `timeupdate` listener immediately and just leave it on. If we add the listener every time we start playing (which is what it seems like from the snippet), we'll eventually end up with a ton of copies of the listener (which won't break anything functionally but will get progressively slower over time).

demos/portal-embed-demo/public/js/ttt/portals-controller.js
```diff
@@ -0,0 +1,214 @@
...
+            const option = {
+                initialY: initialY,
+                initialWidth: initialWidth
+            }
+            this.setDisplayMode(this.PREDECESSOR_ACTIVATE, option);
+        })
+
+        // Display Mode: RESET && EMEBED on predecessor activate
+        this.heroImg.addEventListener('transitionend', evt => {
+            if (evt.propertyName === 'top') {
```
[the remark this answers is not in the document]
uskay: Yes! Thank you for pointing out.
```javascript
            if (evt.propertyName !== 'top') {
                return;
            }
```

demos/portal-embed-demo/public/js/ttt/portals-controller.js
uskay: Split to `setDefaultUI`, `setEmbedUI`, `setPortalActivatedUI` and `setPredecessorActivateUI`.

demos/portal-embed-demo/public/js/ttt/portals-controller.js
```diff
@@ -0,0 +1,214 @@
...
+        // Add event listeners
+        this._hookEvents();
+    }
+
+    /**
+     * Controll all the UI/animations with given display mode
+     * @param {String} mode 
+     * @param {Object} option 
+     */
+    setDisplayMode(mode, option) {
```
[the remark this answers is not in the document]
uskay: Split to `setDefaultUI`, `setEmbedUI`, `setPortalActivatedUI` and `setPredecessorActivateUI`.

demos/portal-embed-demo/public/js/ttt/portals-controller.js
```diff
@@ -0,0 +1,214 @@
...
+                followed: evt.data.followed,
+                name: evt.data.name,
+                photoSrc: evt.data.photoSrc,
+                activatedWidth: evt.data.activatedWidth,
+                predecessor: evt.adoptPredecessor()
+            }
+            // animate the audio controller
+            this.audioController.show();
+            this.audioController.handleActivation();
+            this.setDisplayMode(this.ACTIVATE, option);
```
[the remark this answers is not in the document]
uskay: Split to `setDefaultUI`, `setEmbedUI`, `setPortalActivatedUI` and `setPredecessorActivateUI`.

demos/portal-embed-demo/public/js/ttt/portals-controller.js
```diff
@@ -0,0 +1,214 @@
+/**
+ * Predecessor Controller
+ * Controlling all the portal related features
+ */
+class PortalsController {
```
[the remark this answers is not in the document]
uskay: This could be very very unthoughtful but renamed it to `TTTController`... since it controls everything.

demos/portal-embed-demo/public/js/ttt/ttt-reco.js
uskay: Fixed it!

demos/portal-embed-demo/public/js/ttt/writer-follow.js
uskay: Fixed it!

demos/portal-embed-demo/server.js
uskay: Formatted to 4sp. Used the VS Code formatter.

demos/portal-embed-demo/public/js/portalog/portals-controller.js
uskay: Fixed it!

demos/portal-embed-demo/public/js/ttt/ttt-reco.js
```diff
@@ -0,0 +1,169 @@
...
+     */
+    _createItemTemplate(key, imgURL, title) {
+        return `
+            <div class='pod'>
+                <div class='thumbnail'>
+                    <img src=${imgURL}>
+                </div>
+                <div class='description'>
+                    <div class='title'>${title}</div>
+                    <div class='series'>Tottaly Tooling Tips</div>
```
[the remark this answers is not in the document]
uskay: Thank you for this!

uskay:
Thank you so much for the review. I fixed the code & replied to your comments/suggestions. I've also updated the README.md's demo code reference links (I've test it out. should be working.).

uskay:
Thank you! I'm happy to merge the current version. Once done, I'll link it & finish up the `web.dev` article!

jeremyroman:
Awesome. Thanks so much!

PR #107, opened 2019-05-02
Fixing bugs (README.md and protals-controller.js)

uskay:
Fixing bugs
- Removing "When the promise resolves, it means the page was adopted as a predecessor" comment from README.md (degradation)
- Changing using `append` from `appendChild` for adding fallback elements when the browser does not support Portals. The current implementation is using `appendChild` with multiple nodes as a parameter but it should be `append` in order to do that.
   - Checked other places using `appendChild` to see similar bugs. This one was the only one.


uskay:
@jeremyroman Sorry, 2 bugs I found before publishing `web.dev`.

PR #131, opened 2019-05-17
Adjust behavior around closing and discarding guest browsing contexts.

### LINKED ISSUE #68: "The user agent should not refuse to allow the document to be unloaded."
It's not clear what the intention is here. My most literal reading is that the UA should proceed with all the steps of https://html.spec.whatwg.org/multipage/browsing-the-web.html#prompt-to-unload-a-document, but then even if the user clicks "No don't unload", ignore what the user says.

It seems like a better experience to recommend would be to avoid asking the user entirely. In that case, the operative sentence you want to influence from the "prompt to unload" algorithm is

> then the user agent may ask the user to confirm that they wish to unload the document.

I recognize there is no nice dfn to link to there, but just saying "While executing the [prompt to unload] algorithm, the user agent should not ask the user to confirm that they wish to unload the document" seems good to me.

We could try to make this even more precise, such as adding a flag which we thread all the way through from the "close" algorithm to that "prompt to unload" step. But as long as it stays a "should" that overrides a "may", I think that level of explicitness is not necessary. If we wanted it to be a "must", I'd reconsider.

jeremyroman:
Resolves #68, and attempts to be consistent with iframes by not filing the unload family of events in the event that the context is to be discarded due to removal from the DOM (as opposed to navigation to an empty URL or activation).

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/portals/pull/131.html" title="Last updated on May 24, 2019, 3:38 PM UTC (7819875)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/131/cebe701...jeremyroman:7819875.html" title="Last updated on May 24, 2019, 3:38 PM UTC (7819875)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -28,7 +28,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
             text: document browsing context; url: concept-document-bc
             text: unit of related browsing contexts; url: unit-of-related-browsing-contexts
         urlPrefix: browsing-the-web.html
-            text: refuse to allow the document to be unloaded; url: refused-to-allow-the-document-to-be-unloaded
+            text: prompt to unload; url: prompt-to-unload-a-document
         urlPrefix: history.html
             text: session history; url: session-history
         urlPrefix: infrastructure.html
@@ -42,6 +42,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
             text: valid non-empty URL potentially surrounded by spaces; url: valid-non-empty-url-potentially-surrounded-by-spaces
         urlPrefix: window-object.html
             text: close a browsing context; url: close-a-browsing-context
+            text: discard a browsing context; url: a-browsing-context-is-discarded
 spec: ecma-262; urlPrefix: http://tc39.github.io/ecma262/
     type: dfn
         text: promise; url: sec-promise-objects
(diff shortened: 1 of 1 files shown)


jeremyroman:
@domenic ptal

domenic:
Hmm, I think changing the src="" is more like removing it from the document than it is like the user clicking the close button, so I might use discard there.

jeremyroman:
As I commented in #68, I kinda feel like assigning `src=""` ought to behave similarly to `src="https://some.new.site/"`, which would fire `beforeunload` and `unload`. But if you would rather it not, I can make such a change.

index.bs
```diff
@@ -398,6 +398,17 @@ spec:url; type:dfn; text:scheme
     1. Let |resource| be a new [=request=] whose [=request URL|URL=] is |url| and whose [=request referrer policy|referrer policy=] is "{{no-referrer}}".
 
     1. [=Navigate=] the [=guest browsing context=] to |resource|.
+
+    <div class="note">
+      Unlike an <{iframe}> element, a <{portal}> element supports a state where
+      it has no associated browsing context. This is the initial state of a
+      <{portal}> element (i.e., it has no initial `about:blank` document, but
+      navigates directly to the first legal URL assigned to it).
```
domenic: Nit: `document, but navigates` -> `document; instead it navigates`

Also maybe legal -> parseable

index.bs
```diff
@@ -398,6 +398,17 @@ spec:url; type:dfn; text:scheme
...
+
+      Similarly, a <{portal}> element responds to an illegal <{portal/src}> URL
```
domenic: illegal -> non-parseable

index.bs
```diff
@@ -398,6 +398,17 @@ spec:url; type:dfn; text:scheme
...
+      by closing its browsing context, rather than by navigating to
```
domenic: xref close?

index.bs
jeremyroman: Done, and changed to "parsable", which Wiktionary tells me is preferred for formal writing. (And what are we if not formal?)

index.bs
jeremyroman: Done (unparsable).

index.bs
jeremyroman: Done

jeremyroman:
@lucasgadani ptal

PR #132, opened 2019-05-22
Consistently refer to the guest browsing context as either a browsing context or null.

### LINKED ISSUE #76: "guest browsing context" should be null or a BC
We made similar fixes recently to HTML's notion of "nested browsing context": https://html.spec.whatwg.org/#nested-browsing-context

Then you can say "is null" instead of "If no such context exists", or "set to null" instead of "clear".

jeremyroman:
Resolves #76.

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/portals/pull/132.html" title="Last updated on May 24, 2019, 3:26 PM UTC (b87e976)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/132/2195b2d...jeremyroman:b87e976.html" title="Last updated on May 24, 2019, 3:26 PM UTC (b87e976)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -443,7 +443,7 @@ spec:url; type:dfn; text:scheme
 
   Whenever a <{portal}> element |element| [=becomes browsing-context disconnected=], run the following steps:
 
-  1. If |element| does not have a [=guest browsing context=], abort these steps.
+  1. If |element|'s [=guest browsing context=] is null, abort these steps.
 
   1. [=close a portal element|Close=] |element|.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -260,7 +260,7 @@ spec:url; type:dfn; text:scheme
 
     1. Let |portalBrowsingContext| be the [=guest browsing context=] of [=this=].
 
-        If no such context exists, throw an "{{InvalidStateError}}" {{DOMException}}.
+        If it is null, throw an "{{InvalidStateError}}" {{DOMException}}.
```
domenic: Right below this, you have a similar "if no such context exists" for document browsing context, which has these days also moved to a null-or-BC model. Might as well fix while here.

domenic: The various null check steps now seem better as separate steps, instead of as two paragraphs in one step. But, that's a stylistic choice, up to you.

jeremyroman: Done.

jeremyroman: Done.

jeremyroman:
@lucasgadani ptal

PR #134, opened 2019-05-24
Add a code example to resolve #70.

### LINKED ISSUE #70: Unclear what "complete portal activation" vs. "set the portal state" note is referring to
> Since the task to set the portal state, and thus expose the PortalHost object, is queued before the task to complete portal activation, from the same task source, it is exposed at the time the promise returned from activate() is resolved. 

I can't find any algorithm which both sets the portal state and calls "complete portal activation".

My best guess is that this is referring to a specific user code workflow that calls adoptPredecessor() from the activate event. If so it'd be best to include that code example explicitly. If that's not it though, I'm lost.

jeremyroman:
***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/portals/pull/134.html" title="Last updated on May 24, 2019, 4:00 PM UTC (ba28f5b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/134/e0d59c4...jeremyroman:ba28f5b.html" title="Last updated on May 24, 2019, 4:00 PM UTC (ba28f5b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -206,8 +206,23 @@ spec:url; type:dfn; text:scheme
   <div class="note">
     Since the task to set the [=portal state=], and thus expose the
     {{PortalHost}} object, is queued first, and from the same [=task source=],
-    it is exposed at the time the promise returned from
+    it is exposed at the time the [=PortalActivateEvent/activation promise=] returned from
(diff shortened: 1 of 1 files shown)


jeremyroman:
@domenic

PR #135, opened 2019-05-24
Remove note that repeats that portal browsing contexts are top-level.

### LINKED ISSUE #63: "This implies that every portal browsing context is a top-level browsing context."
In context:

> A portal browsing context has a host which is embeds its rendered output and receives messages sent from the portal browsing context. This specification defines HTMLPortalElement, which hosts a portal browsing context inside an HTML document. A host defines steps to accept a message posted to the host.
>
> This implies that every portal browsing context is a top-level browsing context. It is expected that a Web browser will not present a tab or window to display a portal browsing context, but rather that it will be presented only through a host element. 

It is not clear why the "every portal browsing context is a top-level browsing context" statement follows from the previous paragraph. It seems like that statement is true by design, but it's presumably enforced by the "portal state" transitions only being done on top-level BCs, instead of flowing from the previous paragraph.

The fix here may be to just delete that statement. It's already pretty clear from the opening paragraph of the Concepts section, which states

> Every browsing context has a portal state, which may be none (the default), portal or orphaned. A nested browsing context always has the portal state none.

jeremyroman:
Resolves #63.

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/portals/pull/135.html" title="Last updated on May 24, 2019, 6:31 PM UTC (e702e87)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/135/d297d7d...jeremyroman:e702e87.html" title="Last updated on May 24, 2019, 6:31 PM UTC (e702e87)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -103,12 +103,6 @@ spec:url; type:dfn; text:scheme
   The <dfn>host browsing context</dfn> of a [=portal browsing context=] is its
   [=host element=]'s [=node document|document=]'s [=browsing context=].
 
-  <p class="note">
-    This implies that every [=portal browsing context=] is a [=top-level browsing context=].
-    It is expected that a Web browser will not present a tab or window to display a portal browsing
-    context, but rather that it will be presented only through a [=host element=].
-  </p>
-
   The <dfn>portal task source</dfn> is a [=task source=] used for tasks related to the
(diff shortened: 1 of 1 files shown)


jeremyroman:
@domenic

Issue #123, opened 2019-05-13
Design issues with the <portal> element


ojhunt:
Ok, so it has been requested that I file a bug detailing the issues with <portal> and I don't currently have TAG access. 

I recall having a few more issues, but can't recall what they were at this point.

So here goes:

Usability:
* The primary design goal is to allow "expanding an active frame to a standalone window".  It is unclear how this should interact with the very common user desire to open what are functionally links in a separate tab or window. The specification does not appear to address this behavior. What happens in the context of the host window when a portal is lifted?
  - There's a reference in the spec identifying the obvious problem of how back/forward should work. What happens when the portal window is closed?
  - What happens when I refresh an expanded portal now that I can keep the original window open
* Users are increasingly expecting smooth (eg high frame rate) animation, yet is unclear how that is expected in the context of "animated" expansion of complex layout. Varying behaviour is generally bad UX.

API:
* By creating a separate element the API ends up duplicating a wide array of already specified behaviour. The spec already notes that it needs to specify how errors in postMessage (duplicated API) should be handled - this is already specified and well understood on iframe.

Security:
* There is a major problem in the security properties of <portal> - specifically <iframe>s have historically needed many changes to address security+privacy bugs that are discovered over time. By making <portal> a distinct element rather than an attribute/API on <iframe> we introduce a new spec that must be continuously updated to track the <iframe> specification. This is to my mind a fundamental flaw - many web facing specifications have been discovered to be attackable by existing/known flaws that have been "fixed", but are subsequently found to be duplicated in a separate spec.
* Now that an iframe can be arbitrarily opened we are introducing another clickjacking mechanism - specifically 

Privacy:
* The need for referrer headings is generally considered a privacy flaw, and if we are willing to create an entirely new element we should not reproduce those errors. That means no referrer should be specified or provided in any way.
* How does this interact with various tracker blocking mechanisms? The design goal of any completely new element/API that allows network loads should be to actively prevent their abuse for tracking. Any tracking mechanisms is fundamentally user hostile.
* Historically opening a link doesn't maintain state from the original host environment, this API allows page state to be maintained - do we want to allow this? Now for a user to break the relationship between a "new" window and the origin window.

lucasgadani:
Thanks for the feedback. I'll try to address your points below.

> Ok, so it has been requested that I file a bug detailing the issues with and I don't currently have TAG access.
> 
> I recall having a few more issues, but can't recall what they were at this point.
> 
> So here goes:
> 
> Usability:
> 
[...]

PR #136, opened 2019-05-27
Support referrerpolicy content attribute on <portal>.

### LINKED ISSUE #78: Add referrerpolicy="" attribute?
The spec currently always sets no-referrer, but has a note of

>  Is no-referrer the right referrer policy here, or should we send the document’s URL? 

Is there a reason portals are special, or should we just follow the prevailing pattern for request-creating elements, and add a referrerpolicy="" content attribute which follows [the usual rules](https://html.spec.whatwg.org/#referrer-policy-attribute)?

jeremyroman:
Resolves #78.

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/portals/pull/136.html" title="Last updated on May 27, 2019, 7:43 PM UTC (35d9ef2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/136/8070b25...jeremyroman:35d9ef2.html" title="Last updated on May 27, 2019, 7:43 PM UTC (35d9ef2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -29,6 +29,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
             text: unit of related browsing contexts; url: unit-of-related-browsing-contexts
         urlPrefix: browsing-the-web.html
             text: prompt to unload; url: prompt-to-unload-a-document
+        urlPrefix: common-dom-interfaces.html
+            text: limited to only known values; url: limited-to-only-known-values
         urlPrefix: history.html
             text: session history; url: session-history
         urlPrefix: infrastructure.html
@@ -241,6 +243,10 @@ spec:url; type:dfn; text:scheme
   page that the [=guest browsing context=] is to contain. The attribute, if
   present, must be a [=valid non-empty URL potentially surrounded by spaces=].
 
+  The <dfn element-attr for="portal">referrerpolicy</dfn> attribute is a [=referrer policy attribute=].
+  Its purpose is to set the [=/referrer policy=] used when
+  [=set the source URL of a portal element|setting the source URL of a portal element=]. [[REFERRER-POLICY]]
+
   <p class="note">
     A <{portal}> is similar to an <{iframe}>, in that it allows another browsing context to be embedded.
     However, the [=portal browsing context=] hosted by a <{portal}> is part of a separate [=unit of related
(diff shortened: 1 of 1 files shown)


jeremyroman:
@domenic This is what you're asking for, right?

index.bs
```diff
@@ -268,6 +276,8 @@ spec:url; type:dfn; text:scheme
 
   The <dfn attribute for="HTMLPortalElement">src</dfn> IDL attribute must [=reflect=] the <{portal/src}> content attribute.
 
+  The <dfn attribute for="HTMLPortalElement">referrerPolicy</dfn> IDL attribute must [=reflect=] the <{portal/referrerpolicy}> content attribute, [=limited to only known values=].
```
lucasgadani: Should we specify the empty string defaulting to noreferrer for portals? I think it makes more sense to make the referrer explict, and the spec https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-empty-string says that the empty string falls back to a referrer policy defined elsewhere.

jeremyroman: The precedent everywhere else is that the source browsing context (the one that requested the fetch/subresource)'s referrer policy is used in such cases. It's written in a way that is rather implicit.

That does mean that this is a behavior change, since if the outer document's policy is not no-referrer and the attribute is omitted, we will start sending the referrer.

lucasgadani: While that would be a behavior change, I think it's a positive change form a privacy standpoint. By making the referrer policy more explicit, it gives us privacy by default unless otherwise specified. Given that we don't need to support legacy applications, we should take the opportunity to improve this behavior. As another precedent, we also only support use noopener.

If we are to change that due to ongoing API feedback, it would be much easier to change from disallowed by default than otherwise.

This was also raised in issue #123.

jeremyroman: I'm kinda on the fence about it. We *definitely* should make it easy to send no referrer, in any case. If we want to allow sending a referrer (especially in the 1P case it doesn't seem unreasonable to want it), putting the policy in the hands of the author is the consistent way to do that.

It might be surprising if `<portal>` were the unique element that doesn't respect ambient policy (but only the policy on the element itself, defaulting to `no-referrer`); I'm kinda inclined to defer to @domenic's expertise on the relative costs of a more restrictive default vs a more consistent default.

jeremyroman: `noopener` is slightly different as we have other technical reasons to want to avoid exposing the opener `Window`. :)

lucasgadani: SGTM. I'll defer to @domenic's expertise here on the best behavior to standardize/engage with other browsers.

jeremyroman:
It's also plausible to me that if picking a new default we might want either `no-referrer` or `same-origin`. Filed w3c/webappsec-referrer-policy#121.

@lucasgadani ok to merge as-is, then?

PR #138, opened 2019-05-29
Tie BC creation to availability of a valid src + BC connectedness.

### LINKED ISSUE #137: Allow creation of the portal's browsing context when the src attribute is set
Whenever the src attribute is set, if the guest browsing context is null, we abort the steps. We should consider creating the portal browsing context when the src attribute is set, whenever the element is browsing-context connected.

jeremyroman:
Resolves #137.

PTAL and let me know if you agree that this is the right behavior.

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/portals/pull/138.html" title="Last updated on Jun 20, 2019, 8:39 PM UTC (cc1abcb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/138/61494cd...jeremyroman:cc1abcb.html" title="Last updated on Jun 20, 2019, 8:39 PM UTC (cc1abcb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -441,6 +441,18 @@ spec:url; type:dfn; text:scheme
   <section algorithm="htmlportalelement-may-have-guest-browsing-context">
     To determine whether a <{portal}> element <dfn for="HTMLPortalElement">may have a guest browsing context</dfn>, run the following steps:
 
+    1. If |element|'s [=node document|document=]'s [=browsing context=] is a [=nested browsing context=], then return false.
+
+        <wpt>
+          portals-nested.html
+        </wpt>
+
+        <p class="note">
+          The user agent may choose to emit a warning if the author attempts to
+          use a <{portal}> element in a nested browsing context, as this is not
+          supported.
+        </p>
+
     1. If |element| is [=browsing-context connected=], then return true.
 
     1. If |element|'s [=just-adopted flag=] is true, then return true.
@@ -465,23 +477,7 @@ spec:url; type:dfn; text:scheme
 
     1. Let |hostBrowsingContext| be |element|'s [=node document|document=]'s [=browsing context=].
 
-    1. [=Assert=]: |hostBrowsingContext| is not null.
-
-    1. If |hostBrowsingContext| is a [=nested browsing context=], then:
-
-        1. [=Assert=]: |element|'s [=guest browsing context=] is null.
-
-        1. Return.
-
-        <wpt>
-          portals-nested.html
-        </wpt>
-
-        <p class="note">
-          The user agent may choose to emit a warning if the author attempts to
-          use a <{portal}> element in a nested browsing context, as this is not
-          supported.
-        </p>
+    1. [=Assert=]: |hostBrowsingContext| is a [=top-level browsing context=].
 
     1. If |element| has no <{portal/src}> attribute specified, or its value is the empty string,
         then [=close a portal element|close=] |element| and return.
@@ -539,7 +535,7 @@ spec:url; type:dfn; text:scheme
 
   Whenever a <{portal}> element |element| [=becomes browsing-context connected=], run the following steps:
 
-  1. [=Assert=]: |element| [=may have a guest browsing context=].
+  1. If |element| [=may have a guest browsing context|may not have a guest browsing context=], then abort these steps.
 
   1. If |element|'s [=guest browsing context=] is not null, then abort these steps.
 
--- a/index.bs
+++ b/index.bs
@@ -123,6 +123,15 @@ spec:url; type:dfn; text:scheme
 
     1. [=Assert=]: The [=portal state=] of |predecessorBrowsingContext| is "`none`".
 
+    1. Set the [=host element=] of |successorBrowsingContext| to null.
+
+        User agents *should*, however, attempt to preserve the rendering of the
+        guest browsing context until |predecessorBrowsingContext| has been replaced
+        with |portalBrowsingContext| in the rendering.
+
+        Note: This is intended to avoid a visual glitch, such as a "white flash", where
+        the guest browsing context briefly disappears.
+
     1. Set the [=portal state=] of |predecessorBrowsingContext| to "`orphaned`".
 
     1. Update the user interface to replace |predecessorBrowsingContext| with |successorBrowsingContext|
@@ -329,14 +338,6 @@ spec:url; type:dfn; text:scheme
         |options|["{{PortalActivateOptions/transfer}}"]).
         Rethrow any exceptions.
 
-    1. Set the [=guest browsing context=] of [=this=] to null.
-        User agents *should*, however, attempt to preserve the rendering of the
-        guest browsing context until |predecessorBrowsingContext| has been replaced
-        with |portalBrowsingContext| in the rendering.
-
-        Note: This is intended to avoid a visual glitch, such as a "white flash", where
-        the guest browsing context briefly disappears.
-
     1. Let |promise| be a new [=promise=].
 
     1. Run the steps to [=activate a portal browsing context|activate=] |portalBrowsingContext|
@@ -439,8 +440,6 @@ spec:url; type:dfn; text:scheme
         The user agent *should not* ask the user for confirmation during the
         [=prompt to unload=] step (and so the browsing context should be
         [=discard a browsing context|discarded=]).
-
-    1. Set |element|'s [=guest browsing context=] to null.
   </section>
 
   <section algorithm="htmlportalelement-setsourceurl">
@@ -482,8 +481,6 @@ spec:url; type:dfn; text:scheme
 
         1. Set the [=portal state=] of |newBrowsingContext| to "`portal`", and set
             the [=host element=] of |newBrowsingContext| to |element|.
-
-        1. Set |element|'s [=guest browsing context=] to |newBrowsingContext|.
 
     1. Let |guestBrowsingContext| be |element|'s [=guest browsing context=].
 
@@ -758,11 +755,9 @@ spec:url; type:dfn; text:scheme
   <section algorithm="portalactivateevent-adoptpredecessor">
     The <dfn method for="PortalActivateEvent"><code>adoptPredecessor()</code></dfn> method *must* run these steps:
 
+    1. If [=this=]'s [=PortalActivateEvent/adopted predecessor element=] is not null, throw an "{{InvalidStateError}}" {{DOMException}}.
+
     1. Let |predecessorBrowsingContext| be [=this=]'s [=PortalActivateEvent/predecessor browsing context=].
-
-    1. If |predecessorBrowsingContext| is null, throw an "{{InvalidStateError}}" {{DOMException}}.
-
-    1. Clear [=this=]'s [=PortalActivateEvent/predecessor browsing context=].
 
     1. Let |successorWindow| be [=this=]'s [=PortalActivateEvent/successor window=].
 
--- a/index.bs
+++ b/index.bs
@@ -103,6 +103,13 @@ spec:url; type:dfn; text:scheme
   element which embeds its rendered output and receives messages sent from the
   portal browsing context.
 
+  <div class="note">
+    A <{portal}> element may only be a [=host element=] while it is
+    [=browsing-context connected=] or during the dispatch of the
+    {{Window/portalactivate!!event}} event from which it was obtained
+    using {{PortalActivateEvent/adoptPredecessor()}}.
+  </div>
+
   The <dfn>host browsing context</dfn> of a [=portal browsing context=] is its
   [=host element=]'s [=node document|document=]'s [=browsing context=].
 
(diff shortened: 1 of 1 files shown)


jeremyroman:
Hmm, on further thought, what should happen if you set `src` on a portal that has a guest BC but is not yet inserted? I *think* it should navigate, but it's kinda weird regardless.

```js
onportalactivate = e => {
  let portal = e.adoptPredecessor();
  portal.src = 'https://somewhere.else/';
  /* ... */
};
```

jeremyroman:
Doing some more edge-case thinking here. What about this:

```js
onportalactivate = e => {
  let portal = e.adoptPredecessor();

  // Does this drop the existing BC?
  portal.src = '';

  // If so, does this recreate it, or did the above forfeit the special status?
  portal.src = 'https://somewhere.else/';
}
```

I'm leaning toward making the concept of specially-extended lifetime for the result of `adoptPredecessor` more explicit.

I've been avoiding it because it's kind of indirect, but maybe it's semantically nicer. Something like:

> A \<portal\> element has **lifetime reasons**, which is an ordered set of strings that is initially « ».
>
> To extend *element*'s lifetime due to *reason*, append *reason* to *element*'s lifetime reasons.
> To discontinue *element*'s lifetime due to *reason*, remove *reason* from *element*'s lifetime reasons. If this causes *element*'s lifetime reasons to become empty, then discard *element*'s guest browsing context and set it to null.
>
> Whenever a portal element *element* becomes browsing-context connected, run the following steps:
> 1. Extend *element*'s lifetime due to `"browsing-context-connected"`.
> 1. (Set the source URL if appropriate, etc.)
>
> Whenever a portal element *element* becomes browsing-context disconnected, run the following steps:
> 1. Discontinue *element*'s lifetime due to `"browsing-context-connected"`.

Plus wording to give an additional reason, say `"predecessor-just-adopted"`, in `adoptPredecessor`, which is discontinued at the end of the steps to **activate a portal browsing context**.

This would imply that for the duration of the `portalactivate` event, you can treat the portal element the same as if it were BC-connected (you can set its source to invalid and invalid values, post messages, and so on). I think that's probably nicer than having the two cases be different, even if it seems to demand a little abstraction overhead.

WDYT (before I go to the effort of making such changes)?

@domenic might also want to comment since I'm kinda proposing to make this more complicated in the name of Internal Consistency™.

lucasgadani:
> Doing some more edge-case thinking here. What about this:
> 
> ```js
> onportalactivate = e => {
>   let portal = e.adoptPredecessor();
> 
>   // Does this drop the existing BC?
>   portal.src = '';
> ```

I'd argue that yes, this does drop the BC.

> ```
>   // If so, does this recreate it, or did the above forfeit the special status?
>   portal.src = 'https://somewhere.else/';
> }
> ```

No, because at this point the portal isn't BC-connected. The special status is only achievable by calling the adoptPredecessor API, once it's dropped, it's no possible to get it back.

> I'm leaning toward making the concept of specially-extended lifetime for the result of `adoptPredecessor` more explicit.
> 
> I've been avoiding it because it's kind of indirect, but maybe it's semantically nicer. Something like:
> 
> > A <portal> element has **lifetime reasons**, which is an ordered set of strings that is initially « ».
> > To extend _element_'s lifetime due to _reason_, append _reason_ to _element_'s lifetime reasons.
> > To discontinue _element_'s lifetime due to _reason_, remove _reason_ from _element_'s lifetime reasons. If this causes _element_'s lifetime reasons to become empty, then discard _element_'s guest browsing context and set it to null.
> > Whenever a portal element _element_ becomes browsing-context connected, run the following steps:
> > 
> > 1. Extend _element_'s lifetime due to `"browsing-context-connected"`.
> > 2. (Set the source URL if appropriate, etc.)
> > 
> > Whenever a portal element _element_ becomes browsing-context disconnected, run the following steps:
> > 
> > 1. Discontinue _element_'s lifetime due to `"browsing-context-connected"`.
> 
> Plus wording to give an additional reason, say `"predecessor-just-adopted"`, in `adoptPredecessor`, which is discontinued at the end of the steps to **activate a portal browsing context**.
> 
> This would imply that for the duration of the `portalactivate` event, you can treat the portal element the same as if it were BC-connected (you can set its source to invalid and invalid values, post messages, and so on). I think that's probably nicer than having the two cases be different, even if it seems to demand a little abstraction overhead.
> 
> WDYT (before I go to the effort of making such changes)?

Overall, I think this edge case is so unlikely that it is not worth to overcomplicate the spec to accommodate it in a nicer way. I think it's reasonable to just say that the only way for the document to create the portal browsing context is by appending it to the dom and setting the source.

> @domenic might also want to comment since I'm kinda proposing to make this more complicated in the name of Internal Consistency™.

I'd also like to see @domenic's opinion here, but I'd prefer to keep it simple unless there's an actual use-case that would be solved by this. In fact, even if we left it unspecified, it probably still wouldn't matter, since if implementations disagree on what to do when setting the src immediately after adopting wouldn't be an issue.

jeremyroman:
I would definitely like to not leave it unspecified; I think there's no reason not to want implementations to be compatible about what these operations do and what their preconditions are.

> The special status is only achievable by calling the adoptPredecessor API, once it's dropped, it's no possible to get it back.

I'm not claiming it can be recovered, I'm wondering whether setting `src` forfeited it early. The special status is ordinarily forfeited at the end of the `portalactivate` event.

My aim is simplicity. A more concise way of describing what I wrote above is: a portal may have a BC while it is **browsing-context connected** or it is **the result of adoptPredecessor during portalactivate**. While that is true, it is functional; when it becomes false, it dies.

I think that's simpler than needing to specify the behavior of all of the operations (set src, activate, post message) while in the adopted-but-not-connected state as distinct from both the connected state and the dead state.

jeremyroman:
For comparison: the alternative model (closer to what's currently written) is to predicate operations on the existence of an existing associated browsing context, and if the special state of having-a-guest-but-not-being-connected is lost then the portal element becomes inert until it becomes connected.

index.bs
```diff
@@ -394,9 +394,19 @@ spec:url; type:dfn; text:scheme
...
+    1. If |hostBrowsingContext| is a nested browsing context, then abort these steps.
+
+        <p class="note">
+          The user agent may choose to emit a warning if the author attempts to
+          use a <{portal}> element in a nested browsing context, as this is not
+          supported.
+        </p>
 
     1. If |element| has no <{portal/src}> attribute specified, or its value is the empty string,
         then [=close a portal element|close=] |element| and abort these steps.
```
domenic: Drive-by nit: "return", not "abort these steps", for both.

index.bs
```diff
@@ -407,6 +417,24 @@ spec:url; type:dfn; text:scheme
 
     1. If the [=scheme=] of |url| is not an [=HTTP(S) scheme=], then abort these steps.
 
+        <div class="issue">
+          Would it be more consistent to treat this like an invalid URL, and close?
+        </div>
```
domenic: Making the failure case (including for being in a nested BC) uniformly close seems reasonable to me. Right now there are some cases where the algorithm returns early but doesn't close, which seems like an unnecessary complexity for the author's mental state machine.

index.bs
```diff
@@ -394,9 +394,19 @@ spec:url; type:dfn; text:scheme
   <section algorithm="htmlportalelement-setsourceurl">
     To <dfn for="HTMLPortalElement">set the source URL of a <{portal}> element</dfn> |element|, run the following steps:
```
domenic: "set the source URL" is feeling like a worse name for this algorithm now.

index.bs
jeremyroman: In general for all uses of the phrase "abort these steps", or is there some guidance for when each wording is preferred? If it's just an editorial change in general, maybe I should do a separate PR to globally replace the phrase.

jeremyroman:
I'm not sure "has a BC" can be expressed without reference to state created just for that purpose, because if you invoke `activate`, nothing about its state changes except that it loses its BC. There's a similar case of being adopted by another document, which needs to discard the BC. Forcibly changing content attributes in such a case seems strange.

A weaker "may have a BC" constraint can more easily be described (roughly, BC-connected or in portalactivate dispatch); it could tell you that a `<portal>` definitely has no BC, but not stronger than that.

The lifetime-reasons thing could also be spelled as an algorithm that explicitly checks all of the possible reasons. The `portalactivate` reason would require an explicit boolean of state at least, so it didn't seem shorter. It'd look something like:

> To **check the lifetime** of a `<portal>` element *element*, perform the following steps:
>
> 1. If *element* is browsing-context connected, then return.
> 2. If *element*'s just-adopted flag is true, then return.
> 3. Discard *element*'s guest browsing context.
> 4. Set *element*'s guest browsing context to null.

This would be invoked at all points which could make one of the conditions false (BC disconnection, end of `portalactivate` dispatch).

index.bs
domenic: Yeah, probably best separate. The guidance is not very clear. We use "return" to return from an algorithm, but if that algorithm has sub-steps, e.g. in parallel or a queued task, then we have stuck with "abort these steps" (i.e. "return" is reserved for "exit the whole algorithm").

domenic: In particular, https://infra.spec.whatwg.org/#algorithm-control-flow talks about using return, but it doesn't capture the remaining uses of "abort these steps" that we've evolved. I'll work on clarifying that in Infra.

index.bs
jeremyroman: ok

index.bs
jeremyroman: I'm not sure what a better name is. It's principally used when the embedder directs the portal BC to navigate by changing the `src` attribute. It is also called on BC connection, mainly because we didn't navigate earlier when `src` was set.

"Possibly navigate" sounds a little vague. Any suggestions?

jeremyroman:
Hmm, tried to incorporate some of this feedback into the latest version of this PR. PTAL?

index.bs
```diff
@@ -151,16 +167,24 @@ spec:url; type:dfn; text:scheme
...
-            [=queue a task=] from the [=portal task source=] to the
-            [=event loop=] associated with |predecessorBrowsingContext| to
-            resolve |promise| with undefined.
+        1. Let |adoptedPredecessorElement| be |event|'s [=PortalActivateEvent/adopted predecessor element=].
+
+        1. If |adoptedPredecessorElement| is not null, then:
+
+            1. Set |adoptedPredecessorElement|'s [=just-adopted flag=] to false.
+
+            1. If |element| [=may have a guest browsing context|may not have a guest browsing context=] and its [=guest browsing context=] is not null, then [=discard a browsing context|discard=] it.
```
domenic: The use of discard here vs. close below warrants a note, I think.

index.bs
```diff
@@ -408,25 +440,55 @@ spec:url; type:dfn; text:scheme
...
 
-    1. Let |guestBrowsingContext| be |element|'s [=guest browsing context=].
+    1. [=Assert=]: |element| [=may have a guest browsing context=].
 
-    1. [=Assert=]: |guestBrowsingContext| is not null.
+    1. Let |hostBrowsingContext| be |element|'s [=node document|document=]'s [=browsing context=].
+
+    1. [=Assert=]: |hostBrowsingContext| is not null.
+
+    1. If |hostBrowsingContext| is a nested browsing context, then:
```
domenic:
```suggestion
    1. If |hostBrowsingContext| is a [=nested browsing context=], then:
```

index.bs
```diff
@@ -701,14 +739,17 @@ spec:url; type:dfn; text:scheme
 
   A {{PortalActivateEvent}} has an associated <dfn for="PortalActivateEvent">predecessor browsing context</dfn>,
   which is a [=top-level browsing context=] or null, a <dfn for="PortalActivateEvent">successor window</dfn>, which is
-  a {{Window}}, and an <dfn for="PortalActivateEvent">activation promise</dfn>, which is a [=promise=].
+  a {{Window}}, an <dfn for="PortalActivateEvent">activation promise</dfn>, which is a [=promise=], and a
+  <dfn for="PortalActivateEvent">adopted predecessor element</dfn>, which is a <{portal}> element or null.
```
domenic: The fact that this is an element doesn't appear to be used. A boolean would work just as well. Right?

jeremyroman: It's used to clear the "just adopted" flag after `portalactivate` dispatch.

domenic: You're right, my bad.

index.bs
jeremyroman: Done; wdyt?

index.bs
jeremyroman: Done.

index.bs
```diff
@@ -400,6 +438,16 @@ spec:url; type:dfn; text:scheme
     </wpt>
   </section>
 
+  <section algorithm="htmlportalelement-may-have-guest-browsing-context">
+    To determine whether a <{portal}> element <dfn for="HTMLPortalElement">may have a guest browsing context</dfn>, run the following steps:
+
+    1. If |element| is [=browsing-context connected=], then return true.
```
lucasgadani: Maybe also add "and its browsing context is top-level"?

jeremyroman: Done; PTAL?

Issue #139, opened 2019-05-29
Describe adoptPredecessor use cases in the explainer


jyasskin:
[Some folks think](https://trib.tv/2019/05/28/cake-or-death-amp-and-the-worrying-power-dynamics-of-the-web/) that [`adoptPredecessor()`](https://wicg.github.io/portals/#dom-portalactivateevent-adoptpredecessor) is exclusively useful for UIs like the [AMP carousel](https://developers.google.com/search/docs/guides/about-amp). It'd be helpful for the explainer to describe a couple other use cases.

matthewp:
It's useful in a same-domain context as well. A use case I have in mind is:

A list of images for recipes. Clicking on an image animates to a page for that recipe using a portal. In the recipe page `adoptPredecessor()` is called so that the list page is in a portal. When the user selects a "back" button the reverse of the animation occurs and the list page becomes active again.

twifkak:
I think https://github.com/WICG/portals/issues/114#issuecomment-491409834 applies here.

jyasskin:
i.e. Payment processors! Although that bit of the explainer should include how the [Payment Handler API](https://w3c.github.io/payment-handler/) isn't sufficient.

KenjiBaheux:
Maintain a continuity of the experience between 2 websites operated by the same (or not) company. For an example of the benefit, see [@uskay's demo](https://github.com/WICG/portals/tree/master/demos/portal-embed-demo) which is inspired by a real use case from a partner.

KenjiBaheux:
As [shown at IO](https://youtu.be/rUUazNIZW7I?t=703), keep the context around while performing an action on a third party website. Beside the user friendliness aspect, it can also be used to efficiently perform several actions without having to do a bunch of back forth and scrolling, e.g. save multiple recipes on the spot.

KenjiBaheux:
In the MPA with SPA like transitions demo we did with Hatena (comic viewer), adoptPredecessor could be used to provide an easy way back to the previous chapter for symmetry reasons, instead of using the back button affordance.

jeremyroman:
Hopefully the new changes to `key-scenarios.md` help clarify this.

PR #153, opened 2019-07-03
Add diagrams explaining some additional scenarios where portals could be used.

jeremyroman:
This should address some of the questions from #139.


jeremyroman:
Preview at https://github.com/jeremyroman/portals/blob/addl-key-scenarios/key-scenarios.md

PR #155, opened 2019-08-13
Include a URL change in the retail scenario.

jeremyroman:
URL in the populated details page changes from `retail.example/details` to `retail.example/details/phone` to indicate that the URL has been populated (with `history.replaceState`, presumably).

Requested by Mike Knowles at Google (no linked GitHub account afaict).


jeremyroman:
@mknowles36 ptal?

PR #180, opened 2020-04-24
Meta: minor repository cleanups

domenic:
* Add a .editorconfigs and enforce indents, trailing newlines, and
  trailing whitespace conventions uniformly.
* Add a .gitattributes to get GitHub to display .bs files as HTML.
* Use new-ish gh-pages deploy support to remove the need for deploy.sh.
* Add a Makefile for building the spec
* Fix Bikeshed error (plus one warning) to allow the spec to build again.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/portals/pull/180.html" title="Last updated on Apr 27, 2020, 4:06 PM UTC (3501bfb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/180/40a30b2...3501bfb.html" title="Last updated on Apr 27, 2020, 4:06 PM UTC (3501bfb)">Diff</a>

PR #181, opened 2020-04-24
First draft at new explainer

domenic:
This moves the explainer to the README, and does a quick rewrite pass over the introduction, example, and goals/non-goals, while outlining a skeleton for the rest.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/portals/pull/181.html" title="Last updated on Apr 27, 2020, 6:46 PM UTC (0e830cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/181/29fa541...0e830cd.html" title="Last updated on Apr 27, 2020, 6:46 PM UTC (0e830cd)">Diff</a>


README.md
```diff
@@ -1,6 +1,197 @@
 # Portals
-Portals is a very early proposal for enabling seamless navigations between sites or pages. In particular, this proposal enables a page to show another page as an inset and perform a seamless transition between an inset state and a navigated state.
 
-See the **[explainer](https://github.com/WICG/portals/blob/master/explainer.md)**.
+Portals are a proposed new HTML element which enables seamless navigations between pages. In particular, the `<portal>` element enables a page to show another page as an inset, and then "activate" it to perform a seamless transition to a new state, where the formerly-inset page becomes the top-level document.
 
-See the **[early draft spec](https://wicg.github.io/portals/)**.
+Portals encompass some of the use cases that iframes currently do, but with better security, privacy, and ergonomics properties. And via the activation operation, they enable new use cases like preloading or navigation transitions.
```
domenic, line 5: I wonder if it'd be worth mentioning the "super-links" framing here.

README.md
```diff
@@ -1,6 +1,197 @@
...
+TODO are we still thinking of making this the default click action? If so we should state that here. If it's still under debate we should state that it's under debate and link to an open issue.
+
+At this point, the user will observe that their browser has navigated to `https://www.example.com/` (e.g., via changes to the URL bar contents and back/forward UI). Since `https://example.com/` was already loaded in the portal context, this navigation will occur seamlessly and instantly, without a network round-trip or document re-initialization.
+
+For more advance use cases, the `https://www.example.com/` document can react to activation, using the `portalactivate` event. It can use this event to adapt itself to its new context, and even to adopt its *predecessor* (the document which previously occupied the tab) into a new portal context.
+
+```js
+window.addEventListener('portalactivate', e => {
+  document.body.classList.add('displayed-fully');
+  document.requestStorageAccess();
```
domenic, line 52: I threw in some extra lines here because the code seemed uncompelling (and hopefully unnecessary?) if it was basically just "keep me around to work around the lack of bfcache". Suggestions for better stuff welcome.

README.md
```diff
@@ -1,6 +1,197 @@
...
+
+- **Aggregation**: multiple portals on the same page can be used to create more elaborate experiences, where the user chooses which portal to activate. This category of use cases includes cases like a news reader, a shopping site, an infinite scrolling list of articles, etc. By using a portal instead of (or in addition to) a link, the aggregated content has the opportunity to display a preview, and to benefit from pre-rendering and navigation transitions.
+
+  Additionally, by using the ability to adopt the predecessor during the `portalactivate` event, more complicated interactions between the aggregator and the aggregated content can be built, such as retaining a portion of the shopping site or article-list UI in a portal even after navigating to an individual page.
+
+- **"Better iframe"**: portals encompass some, but not all, of the use cases for iframes. And they do so in a way that is better for users, in terms of security and privacy. They also remove a lot of the legacy baggage and sharp edges that come with iframes, making them easier to use for web developers. So we anticipate many of the current cases for iframes, such as ads, being able to be replaced with portals.
```
domenic, line 74: "such as ads" might not be super-accurate here. In general it'd be good to get a better sense of which iframe use cases can be done via portals. (That's one of the TODOs below.)

README.md
```diff
@@ -1,6 +1,197 @@
 # Portals
-Portals is a very early proposal for enabling seamless navigations between sites or pages. In particular, this proposal enables a page to show another page as an inset and perform a seamless transition between an inset state and a navigated state.
 
-See the **[explainer](https://github.com/WICG/portals/blob/master/explainer.md)**.
+Portals are a proposed new HTML element which enables seamless navigations between pages. In particular, the `<portal>` element enables a page to show another page as an inset, and then "activate" it to perform a seamless transition to a new state, where the formerly-inset page becomes the top-level document.
```
jakearchibald, line 3: > becomes the top-level document.

Does this language work if the portal is activating within an iframe?

README.md
jakearchibald: Given the concessions we need to make for cross-origin portals, I'm a little nervous about about saying they're better than iframes. They comes with additional restrictions compared to iframes, so they can be activated without compromising privacy. But ergonomics wise, things like the rendering style will make them less-useful for some of the things folks use iframes for today.

README.md
```diff
@@ -1,6 +1,197 @@
...
+
+This is somewhat like an iframe, in that it provides a rendering of the specified document at `https://example.com/`. However, a portal is much more restricted than an iframe: user interaction with it is [limited](#interactivity) to just clicks (like a button or link), and capabilities like storage and cross-origin communication are [prohibited](#privacy-threat-model-and-restrictions) while rendering in the portal context. This allows portals to be used as a more-restricted view onto another document when the full capabilities of an iframe are not needed.
+
+In exchange for these restrictions, portals gain an additional capability that iframes do not have. A portal can be *activated*, which causes the embedding window to navigate, replacing its document with the portal:
+
+```js
+myPortal.activate();
+```
+
+TODO are we still thinking of making this the default click action? If so we should state that here. If it's still under debate we should state that it's under debate and link to an open issue.
```
jakearchibald, line 43: It's still under debate.

README.md
```diff
@@ -1,6 +1,197 @@
...
+
+At this point, the user will observe that their browser has navigated to `https://www.example.com/` (e.g., via changes to the URL bar contents and back/forward UI). Since `https://example.com/` was already loaded in the portal context, this navigation will occur seamlessly and instantly, without a network round-trip or document re-initialization.
```
jakearchibald, line 45: > this navigation will occur seamlessly and instantly, without a network round-trip or document re-initialization.

There may be cases where we need to force a reload. Eg, if a GitHub page is portaled, the storage restrictions will make it render in a logged-out experience, at best. Activating the page in this state may make it look like GitHub is broken/unreliable, so I think GitHub would need to acknowledge the change in storage partitioning to avoid a forced reload.

README.md
```diff
@@ -1,6 +1,197 @@
...
+
+For more advance use cases, the `https://www.example.com/` document can react to activation, using the `portalactivate` event. It can use this event to adapt itself to its new context, and even to adopt its *predecessor* (the document which previously occupied the tab) into a new portal context.
```
jakearchibald, line 47:
```suggestion
For more advanced use cases, the `https://www.example.com/` document can react to activation, using the `portalactivate` event. It can use this event to adapt itself to its new context, and even to adopt its *predecessor* (the document which previously occupied the tab) into a new portal context.
```

README.md
jakearchibald: I'm not sure if you'd use a portal for things that don't need to activate.

README.md
```diff
@@ -1,6 +1,197 @@
...
+
+Non-goals:
+
+- Built-in support for high-level navigation patterns, such as carousels or infinite lists. Portals provide a low-level building block for pre-rendering, which can be combined with the usual tools of HTML for creating navigation pattern UIs.
+
+- Built-in support for portal-specific transition animations. Given that portals are represented by HTML elements, existing CSS mechanisms are enough to allow authors to create compelling navigation transitions.
+
+- Subsume all the use cases of iframes. The use cases for portals overlap with those for iframes, but in exchange for the ability to be activated, portaled pages lose abilities like cross-origin communication, storage, and nontrivial interactivity. As such, portals are not suitable for use cases like embedded widgets.
+
+- Allowing arbitrary unmodified web pages to be portaled. Pages will need to adapt to work well when they are hosted in a portal.
```
jakearchibald, line 96: fwiw, I don't think some-origin pages will need to adapt to being portaled.

index.bs
```diff
@@ -885,7 +895,7 @@ spec:url; type:dfn; text:scheme
     The algorithm for determining the [[CSP#effective-directive-for-a-request|effective directive for a request]]
     is modified by prepending the following:
 
-    1. If |request|'s [=initiator=] is "", its [=destination=] is "`document`", and
+    1. If <var ignore>request</var>'s [=initiator=] is "", its [=destination=] is "`document`", and
```
jakearchibald, line 898: I'd leave this as it was, but change the previous line:

> The algorithm for determining the [[CSP#effective-directive-for-a-request|effective directive for a request]], |request| (a [=/request=]) is modified by prepending the following:

README.md
domenic: Right, I tried to be explicit it only covers some of the use cases iframes do, with some more details below.

README.md
domenic: I mean, ads need to be navigated to, and activating them seems like a nicer way of navigating? I'm not sure though.

README.md
domenic: Portals cannot load inside an iframe.

README.md
domenic: Hmm. Would the browser force the reload, or would the page add `location.reload()` in its `portalactivate` handler? If the former, how would we know when to force the reload and when not?

README.md
jakearchibald: Ohh I didn't realise that. Hah.

index.bs
domenic: Good idea. I've done this in https://github.com/WICG/portals/pull/180 and will rebase this on that when it's merged.

README.md
```diff
@@ -1,6 +1,197 @@
...
+### Interactivity
+
+TODO:
+
+- Discuss how portals are only trivially interactive, like buttons or links. Not like iframes.
+- Discuss focus. (Again, like buttons or links.)
+- Did we end up adding a default click behavior like links?
+- Discuss scrolling, including the problem of scroll handoff and its importance. Note that scroll handoff is a ??? in spec terms but we promise to spec something interoperably implementable, somehow.
+
+### Accessbility
```
jyasskin, line 144:
```suggestion
### Accessibility
```

index.bs
```diff
@@ -60,6 +60,15 @@ spec:url; type:dfn; for:/; text:url
 spec:url; type:dfn; text:scheme
 </pre>
 
+<details class="annoying-warning">
+  <summary>This spec is behind the explainer</summary>
+
+  <p>This specification document has not yet been updated to reflect the 2020-04 updates to
+  <a href="https://github.com/WICG/portals">the explainer</a>. We'll fix that as soon as we can, but
```
jyasskin, line 67:
```suggestion
  <a href="https://github.com/WICG/portals#readme">the explainer</a>. We'll fix that as soon as we can, but
```
to scroll straight there.

README.md
```diff
@@ -1,6 +1,197 @@
...
+TODO:
+
+- A new attribute on iframes (text exists in explainer.md but is very implementer-focused)
+- Other (historical?) solutions to prerendering
+- Other (historical?) solutions to navigation transitions
+- Adding activation ("promotion") to iframes (text exists in explainer.md but is very implementer-focused)
+- Using the fullscreen API (text exists in explainer.md but is very implementer-focused)
+- Allowing cross-origin communication and storage
+
+## Acknowledgments
```
jyasskin, line 181:
```suggestion
## Security and privacy considerations

TODO

### Security/Privacy Questionnaire

This section contains answers to the [W3C TAG Security and Privacy Questionnaire](https://w3ctag.github.io/security-questionnaire/).

1. TODO

## Stakeholder Feedback

* W3CTAG: ?
* Browsers:
  * Safari: ?
  * Firefox: ?
  * Samsung: ?
  * UC: ?
  * Opera: ?
  * Edge: ?
* Web developers: ?

## Acknowledgments
```

From https://w3ctag.github.io/explainers#stakeholder-feedback--opposition and general good practice. It's possible the questionnaire should live in a separate document.

README.md
```diff
@@ -1,6 +1,197 @@
...
+
+- Maybe introduce by telling people to think of portals as super-links, not super-iframes. At least from a user's perspective; from the spec and implementer perspective they're more related to iframes, and for web developers they're kind of in between.
+- Reiterate everything from the above sections in one convenient place.
+- Include minor things as well (like, doesn't appear in `window.frames`).
+
+## Alternatives considered
+
+TODO:
+
+- A new attribute on iframes (text exists in explainer.md but is very implementer-focused)
```
jyasskin, line 174: Text doesn't exist there anymore. Maybe
```suggestion
- A new attribute on iframes ([text exists in explainer.md](https://github.com/WICG/portals/blob/d901563053af7d56a2cafee686ee89d651c49eb7/explainer.md#why-do-we-need-a-new-html-tag-why-not-use-iframe) but is very implementer-focused)
```

README.md
```diff
@@ -1,6 +1,197 @@
...
+- Other (historical?) solutions to prerendering
+- Other (historical?) solutions to navigation transitions
+- Adding activation ("promotion") to iframes (text exists in explainer.md but is very implementer-focused)
+- Using the fullscreen API (text exists in explainer.md but is very implementer-focused)
```
jyasskin, line 178: Ditto here.

README.md
domenic: Are there other cases to mention where people might want to use portals instead of iframes? Maybe same-origin ones?

README.md
domenic: Makes sense. I agree we can leave out that level of detail here, although we should have an open issue recording it. Can one of you take point on writing that up in the issue tracker?

README.md
domenic: Done. More ideas still welcome.

README.md
domenic: For these purposes I'll just add "Cross-origin" to the beginning.

README.md
domenic: I think a separate document is probably best for the questionnaire. I'll add a skeleton for it.

README.md
```diff
@@ -1,6 +1,197 @@
...
+
+## Example
+
+A document can include a `<portal>` element, which renders the specified URL in a portal context:
+
+```html
+<portal id="myPortal" src="https://www.example.com/"></portal>
+```
+
+This is somewhat like an iframe, in that it provides a rendering of the specified document at `https://example.com/`. However, a portal is much more restricted than an iframe: user interaction with it is [limited](#interactivity) to just clicks (like a button or link), and capabilities like storage and cross-origin communication are [prohibited](#privacy-threat-model-and-restrictions) while rendering in the portal context. This allows portals to be used as a more-restricted view onto another document when the full capabilities of an iframe are not needed.
```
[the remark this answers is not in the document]
domenic: I rephrased this to be more accurate; PTAL!

README.md
```diff
@@ -1,6 +1,222 @@
...
+
+  let predecessor = e.adoptPredecessor(document);
+  console.assert(predecessor instanceof HTMLPortalElement);
+  document.body.appendChild(predecessor);
+});
+```
+
+## Use cases
+
+_See the ["Key Scenarios"](./key-scenarios.md) document for more detail on each of these, including visualizations._
```
jyasskin, line 66: This contradicts the warning at the top of `key-scenarios.md` a bit, but it's probably worth keeping anyway.

README.md
```diff
@@ -1,6 +1,222 @@
...
+
+- **Pre-rendering**: by loading a page in a hidden portal, it is possible to pre-render the destination page. Then, activating the portal will instantly display the pre-rendered document in the same browsing context.
```
jyasskin, line 68: Is "browsing context" right? Or should this say
```suggestion
- **Pre-rendering**: by loading a page in a hidden portal, it is possible to pre-render the destination page. Then, activating the portal will instantly display the pre-rendered document in the same browser tab.
```
to be more vague about the architecture?

domenic: I think both are right, but I prefer your user-facing language suggestion.

README.md
jeremyroman: @jakearchibald is probably better equipped to articulate the problem, since he was the one who pointed it out originally.

jakearchibald: On it.

jakearchibald: https://github.com/WICG/portals/issues/184

PR #183, opened 2020-04-27
Update Bikeshed to Python 3.

jyasskin:
There are still two errors in the spec that are fixed by #181.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,13 @@
 language: python
 sudo: false
 python:
-  - "2.7"
+  - "3.7"
 
 install:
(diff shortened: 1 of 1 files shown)


domenic:
https://github.com/WICG/portals/pull/180 takes care of this by using the "living Bikeshed" already installed on the CSSWG server, instead of reinstalling Bikeshed on a new VM each build.

jyasskin:
As far as I know, #180 can't be tweaked to enable `--die-on=warning`.

domenic:
Did you see https://github.com/WICG/portals/pull/180/files#diff-b67911656ef5d18c4ae36cb6741b7965R9 ?

jyasskin:
Ah, sorry, never mind then.

PR #189, opened 2020-05-04
Rearrange spec structure as more explicit patches

domenic:
This brings the spec closer to how it would look upon landing in HTML, and explicitly calls out what sections it would expect things to land in. Notable changes:

* Moved window.portalHost from "Miscellaneous extensions" into a new "Portal hosts" section, which contains the definitions for both window.portalHost and the PortalHost interface.

* Made the onportalactivate event handler addition match HTML's existing structure.

* Moved subsections of "Security considerations" which were spec patches into a new dedicated section, "Updates to other specifications".

* Noted the connection to whatwg/html#1230 in addition to RFC 7034.

* Tweaks the wording in the "Fetch Metadata Request Headers" section's non-normative summary notes.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/portals/pull/189.html" title="Last updated on May 4, 2020, 9:36 PM UTC (45881b2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/189/d69147d...45881b2.html" title="Last updated on May 4, 2020, 9:36 PM UTC (45881b2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -907,8 +907,8 @@ spec:url; type:dfn; text:scheme
         <td><dfn event for="Window"><code>portalactivate</code></dfn></td>
         <td>{{PortalActivateEvent}}</td>
         <td>{{Window}}</td>
-        <td>Fired at the {{Window}} when it is associated with a new [=top-level browsing context=]
-        due to activation of its [=portal browsing context=].</td>
+        <td>Fired at the {{Window}} of a [=portal browsing context=] when that [=portal browsing
+        context=] is activated.</td>
       </tr>
     </tbody>
   </table>
--- a/index.bs
+++ b/index.bs
@@ -78,11 +78,17 @@ spec:url; type:dfn; text:scheme
   This specification extends [[HTML]] to define a new kind of [=top-level browsing context=],
   which can be embedded in another document, and a mechanism for replacing the contents of another
   top-level browsing context with the previously embedded context.
+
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -819,76 +860,66 @@ spec:url; type:dfn; text:scheme
...
+        <th>Description</th>
       </tr>
     </thead>
     <tbody>
       <tr>
         <td><dfn event for="Window"><code>portalactivate</code></dfn></td>
         <td>{{PortalActivateEvent}}</td>
-        <td>The window is associated with a new [=top-level browsing context=] due to activation of its [=portal browsing context=].</td>
+        <td>{{Window}}</td>
+        <td>Fired at the {{Window}} when it is associated with a new [=top-level browsing context=]
```
jeremyroman, line 910: Well, the {{Window}} (i.e. the object itself) is still associated with the same top-level browsing context; it's the UA-level window that is associated with a new browsing context. This wording was awkward before this PR too (probably because this was one of the earlier things to be written and terminology has changed).

domenic: Hmm, yeah, I see. Maybe something like this?

> Fired at the Window of a portal browsing context when that portal browsing context is activated.

jeremyroman: Sure, sg. The exact details are elsewhere, so this only needs to give the right general intution.

PR #190, opened 2020-05-04
Expand on the differences between portals and iframes

domenic:
Critiques and suggestions welcome!

I'd also be interested in @jakearchibald's review (I just invited him to the repo so I can formally request it.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@ Portals encompass some of the use cases that iframes currently do, but with bett
   - [Session history, navigation, and bfcache](#session-history-navigation-and-bfcache)
 - [Summary of differences between portals and iframes](#summary-of-differences-between-portals-and-iframes)
 - [Alternatives considered](#alternatives-considered)
+  - [A new attribute on an existing element](#a-new-attribute-on-an-existing-element)
+  - [TODO:](#todo)
 - [Security and privacy considerations](#security-and-privacy-considerations)
 - [Stakeholder feedback](#stakeholder-feedback)
 - [Acknowledgments](#acknowledgments)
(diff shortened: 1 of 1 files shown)


jeremyroman:
lgtm with a caveat that this wording sounds like it explicitly forecloses some possible future directions of evolution (if, e.g., we want a mode with some kind of interactivity enabled in the future; @KenjiBaheux might be best positioned to comment on that)

I do agree that we want a reasonable sharp/clear delineation in our explainer so that it doesn't end up too muddied by all of the things that could possibly be added but don't exist now.

domenic:
Hmm, yeah, it might be worth adding a "Potential future expansions" section in the explainer and then referencing that---while still making the case that the default behavior is quite different. But I'm not sure; we've got enough work ahead of us for explaining "v1".

README.md
```diff
@@ -161,21 +163,56 @@ TODO:
...
+Finally, the web developer dealing with a portal element's API sees the following differences from iframes:
+
+- Even same-origin portals do not provide synchronous DOM access to the portaled `Window` or `Document` objects, whereas iframes give such access via `frame.contentWindow`/`frame.contentDocument`. This gives a more uniform isolation boundary for more predictable performance and security.
+
+- Similarly, portaled `Window` objects are not accessible via accessors like `window.iframeName` or `window[0]`, and they cannot access related `Window` objects via `top` or `parent` (or `opener`).
 
-- Maybe introduce by telling people to think of portals as super-links, not super-iframes. At least from a user's perspective; from the spec and implementer perspective they're more related to iframes, and for web developers they're kind of in between.
-- Reiterate everything from the above sections in one convenient place.
-- Include minor things as well (like, doesn't appear in `window.frames`).
+- Portals cannot be made to navigate in the way iframes (or popups) can, via `window.open(url, iframeName)`.
```
jakearchibald, line 182: Also, portal navigations are replacement, and don't impact the back/forward button of the top level page.

PR #204, opened 2020-05-20
Tweak PortalActivateOptions IDL

### LINKED ISSUE #74: IDL tweaks for post messaging
```webidl
dictionary PortalActivateOptions {
    any data = null;
    sequence<object> transfer = [];
};
```

I think this would be better as

```webidl
dictionary PortalActivateOptions : PostMessageOptions {
    any data;
};
```

(the normative difference being that the default is undefined instead of null.)

Separately I considered whether `message` or `data` was better, but I concluded matching the web-dev-visible property name on `MessageEvent` was more important than matching the non-visible spec name for the parameter to postMessage.

domenic:
The normative change here is that by default data is undefined instead of null. Closes #74.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/portals/pull/204.html" title="Last updated on May 20, 2020, 8:52 PM UTC (cbfeef7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/204/7200658...cbfeef7.html" title="Last updated on May 20, 2020, 8:52 PM UTC (cbfeef7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -333,9 +333,8 @@ spec:url; type:dfn; text:scheme
           attribute EventHandler onmessageerror;
       };
 
-      dictionary PortalActivateOptions {
-          any data = null;
-          sequence<object> transfer = [];
+      dictionary PortalActivateOptions : PostMessageOptions {
+          any data;
       };
   </xmp>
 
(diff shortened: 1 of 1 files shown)


jeremyroman:
lgtm; we should make the equivalent change in chromium (or at least file a bug, but this should be small-ish) so we don't forget

domenic:
Filed https://bugs.chromium.org/p/chromium/issues/detail?id=1085197

Issue #191, opened 2020-05-06
Currently, knowing when a portal can be activated is guesswork


jakearchibald:
From https://github.com/WICG/portals/issues/185, an activation can fail if:

> > > no document was yet loaded in the portal, e.g. because the fetch isn't far enough along to commit, or because the UA chose to block it for some reason (incl. but not limited to CSP)
> > 
> > Do developers have a way to detect when the fetch _is_ far enough along?
> 
> They have a way to detect later events, like load. I don't know whether we can reasonably expose this information. This is necessary because the entire load could end up being blocked and then we have nothing to show the user. This could have already happened, or we could be sitting waiting on a request which will ultimately result in a redirect to an invalid URL, a response with a CSP that prevents loading, or something similar.

This doesn't feel right. The load event is far too long to delay a navigation, and the only other choice we're leaving developers with is taking a punt and calling `activate()` early.

If exposing an earlier point is a leak, then we're kinda already leaking it though `activate()` rejecting.

---

When the developer calls `activate()`, we could wait until the fetch _is_ far enough along, before deciding whether to fulfill or reject. This is generally how promise-based APIs work – they don't fail if the thing is in-progress.

Also, it'd be great to provide a `portal.contentPainted` promise, which is a signal for "you can start the transition now". But yeah, I'm unsure of the security implications here.

jeremyroman:
We could make `activate()` wait, though it's kinda suboptimal if the developer did an animation already. Perhaps better than the current behavior, though. :)

`activate()` rejecting is less of a leak when/if we require a user gesture, though, which we seem likely to.

jakearchibald:
> We could make `activate()` wait, though it's kinda suboptimal if the developer did an animation already. Perhaps better than the current behavior, though. :)

Agreed, this is why I think we need something like `portal.contentPainted` to avoid animating a transparent portal.

jeremyroman:
I do like this idea. I wonder if we'd want it to be a promise or an event. If a promise, I think I would expect it to be replaced with a new promise when I set `src` (but not when navigations occur internally to the portal context), but I haven't given it much thought.

jakearchibald:
> I wonder if we'd want it to be a promise or an event.

Even though it can happen many times, I think a promise is more usable as it can't be 'missed' (see the libraries that try to solve this for DOMContentLoaded).

> I think I would expect it to be replaced with a new promise when I set `src`

Agreed. That was also the conclusion in https://github.com/whatwg/html/issues/127.

[...]

Issue #185, opened 2020-05-04
Document reasons why portal activation can fail


jeremyroman:
Currently the following cases exist in the Chromium implementation. Those marked with an asterisk (*) are already expressed in the draft:

Synchronously:
1. no browsing context associated with the element*
1. another HTMLPortalElement has an outstanding activation
1. the element is itself inside a portal browsing context*
1. activate data serialization failure

Asynchronously:

1. a top-level navigation was in progress when the browser tried to execute the activation
1. no document was yet loaded in the portal, e.g. because the fetch isn't far enough along to commit, or because the UA chose to block it for some reason (incl. but not limited to CSP)
1. an error page (interstitial) is being displayed, rather than the requested resource

Of these, no browsing context and the element itself being inside a portal browsing context are already expressed in the spec, and the remaining five are not.

jeremyroman:
@domenic three of the seven possible reasons are currently expressed in the draft

jakearchibald:
> a top-level navigation was in progress when the browser tried to execute the activation

Is this consistent with other forms of navigation? I thought in-process navigations were cancelled if a new navigation happens, but I'm sure it's more complicated than that.

> no document was yet loaded in the portal, e.g. because the fetch isn't far enough along to commit, or because the UA chose to block it for some reason (incl. but not limited to CSP)

Do developers have a way to detect when the fetch _is_ far enough along?

kjmcnee:
In chromium at least, renderer initiated navigations without a user gesture are not allowed to cancel a browser initiated navigation. The idea is to prevent a page from trapping a user who is trying to use the back button by navigating repeatedly.

We have the same behaviour for portals: a portal activation without a user gesture cannot cancel a browser initiated navigation.

jeremyroman:
> > a top-level navigation was in progress when the browser tried to execute the activation
> 
> Is this consistent with other forms of navigation? I thought in-process navigations were cancelled if a new navigation happens, but I'm sure it's more complicated than that.

Ultimately there's a point at which the browser has decided to go ahead with the navigation. In Chromium, at least, this is triggered by the beginning of the response body. At this point we have changed, or are about to change upon receiving a rendered frame, the page that the user is looking at. Activation cannot proceed at that point.

Navigations are cancelled by new navigations, unless they have reached this point.

> > no document was yet loaded in the portal, e.g. because the fetch isn't far enough along to commit, or because the UA chose to block it for some reason (incl. but not limited to CSP)
> 
[...]

PR #208, opened 2020-06-05
Flesh out the privacy threat model.

jyasskin:
Let me know if there are other side channels we're blocking.


README.md
```diff
@@ -105,16 +105,41 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
 
 ### Privacy threat model and restrictions
 
+The Portals design is intended to comply with the [W3C Target Privacy Threat Model](https://w3cping.github.io/privacy-threat-model/). This section discusses the aspects of that threat model that are particularly relevant to Portals and how the design satisfies them.
+
+A portal can contain either a same-site or cross-site resource. Same-site portals don't present any privacy risks, but cross-site resources risk enabling [cross-site recognition](https://w3cping.github.io/privacy-threat-model/#model-cross-site-recognition) by creating a messaging channel across otherwise-partitioned domains.
```
domenic, line 110: Interesting that this focuses on site over origin, but I guess it makes sense.

My understanding is that we're planning to block various channels for cross-_origin_ portals. This might deserve a sentence or two.

README.md
```diff
@@ -105,16 +105,41 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
...
+
+This design assumes that storage will be [partitioned](https://github.com/privacycg/storage-partitioning), even though that is not the status quo in all browsers. Because portaled documents can be activated into a top-level browsing context, they (eventually) live in the first-party storage partition of their origin. This means we have to prevent communication with the surrounding document to the same extent we prevent it with a cross-site link opened in a new tab.
+
+#### Channels that are blocked
```
domenic, line 114: This section contains both channels that are blocked for both for cross-origin portals (postMessage, credentials/storage, ...) and things which we decided to make the same everywhere (most notably initial size) for simplicity. It's probably worth signaling which category each restriction is in. Maybe with a blanket statement plus a sentence noting that portal resizing/initial sizing is an exception?

README.md
```diff
@@ -105,16 +105,41 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
...
+
+* [`postMessage()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) isn't allowed between a portal and its container.
+* The sequence of portal loads: The source page can encode its user ID into the order in which a sequence of URLs are loaded into portals. To prevent the target from correlating this ID with its own user ID without a navigation, documents loaded into portals are fetched without credentials and don't have access to either first-party storage or multi-keyed storage. They have to use [`requestStorageAccess()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/requestStorageAccess) (or another TBD mechanism) to get access while they're being activated.
+* Side channels:
+  * Portal Resize: Javascript outside the portal could use a sequence of portal resizes to send a user ID, so portals cannot be resized after creation.
```
domenic, line 119: I believe the current model (discussed a bit in https://github.com/WICG/portals#rendering, which might deserve a cross-link) is that the containing page can resize the `<portal>` element as much as it wants, but this just causes the contents to get re-rasterized. The inner page does not know about the resizing; it is always painting into something that is the same size as the top-level tab.

I guess it's kind of unclear though, since that section still has "Discuss viewport size. Full viewport size at all times? Resizing OK or no? I'm sketchy on the plan here." Maybe we should take this opportunity to firm that up.

README.md
```diff
@@ -105,16 +105,41 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
...
+  * Intersection Observer: Won't give visibility information to script inside the portal, to avoid occlusion being used to send information.
+  * [Page Visibility](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) and whether `requestAnimationFrame()` callbacks run, as in normal iframes, match the visibility of the top-level page to prevent the page from encoding messages in visibility changes. However, this leads to the possibility that a portal and containing page could use the timing of user-caused visibility changes to join the user across site boundaries. Whether and how to prevent this is [still under discussion](https://github.com/WICG/portals/issues/193#issuecomment-639768218).
+
+#### Channels with 1 bit per user interaction
+
+* User clicks a link, and the target site decides between a 204 or other non-navigating response, vs a real response, based on the user's id on the target site. This requires [specialized code on the target server](https://w3cping.github.io/privacy-threat-model/#cap-run-on-server), which allows [full user ID transfer](https://w3cping.github.io/privacy-threat-model/#model-cross-site-recognition) by other means.
```
domenic, line 126: How exactly does this leak work? Is it via `load` events? Would it help if we only fired `load` events for the initial portal load, and not for any links being clicked inside the portal?

README.md
```diff
@@ -105,16 +105,41 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
...
+
+#### Channels with a few noisy bits per user interaction
+
+* User clicks a portal, and the target site delays the load by an amount of time that depends on the user's ID. This is noisy because it's hard for the original site to know how much time to expect for that load on this user's machine, but it could encode several bits. This information is also available for normal navigations via the `pagehide` and `beforeunload` events. This requires [specialized code on the target server](https://w3cping.github.io/privacy-threat-model/#cap-run-on-server), which allows [full user ID transfer](https://w3cping.github.io/privacy-threat-model/#model-cross-site-recognition) by other means.
+
+#### Channels that match navigation
+
+* The url of the activated portal and the referring URL are available to normal navigations to the same extent they're available to a portal. Solutions to link decoration will apply to both.
```
domenic, line 134:
```suggestion
* The URL of the activated portal and the referring URL are available to normal navigations to the same extent they're available to a portal. Solutions to link decoration will apply to both.
```

README.md
jyasskin: I'm arguing here that we don't need to fix this leak. But yes, it'd work via any "you can safely activate this portal now" event on the outer portal load. IIUC, you can't click links inside the portal since all interaction is blocked.

README.md
jyasskin: Thanks, done.

README.md
jyasskin: I've mentioned cross-origin or same-origin explicitly in each case, since neither is overwhelmingly common. None of these restrictions are currently planned to just be cross-site, right?

README.md
domenic: > My understanding of the goal is that: an origin's owner can opt into information passing between origins, but the end-user has to opt into information passing between sites. Obviously we're not entirely there yet.

This is probably reasonable, although I'll note that for largely architectural/implementation reasons we're trying to get rid of `document.domain`. (And I guess we're trying to get rid of cookies too?) So there might be a future where the end-user has to opt into information passing between origins.

> Origins are more predictable than sites, so that could be a good reason to block cross-origin channels: to make it easier for developers to understand the restrictions. Is that why we're planning to block the cross-origin channels too, or is there an attack I've missed?

I think that's right. Plus the general program above of slowly trying to get rid of cross-origin same-site information sharing mechanisms means that we don't want to add any _new_ ones.

README.md
domenic: > one of these restrictions are currently planned to just be cross-site, right?

Correct.

README.md
domenic: > I'm arguing here that we don't need to fix this leak. But yes, it'd work via any "you can safely activate this portal now" event on the outer portal load.

Hmm. I don't quite understand then. Could you outline the leak in more detail? What link is being clicked here, and how is that link related to a portal?

> IUC, you can't click links inside the portal since all interaction is blocked.

True. I meant to discuss cases like JavaScript inside the portal doing `location.href = "another-page.html"` and that causes a load event. I believe that causes a load event for iframes.

README.md
domenic: The clarifications here look great, thank you!

README.md
domenic: Your current updates look pretty solid. I'd be comfortable merging as-is and revisiting when we write down the rescaling behavior in more precise detail (in `#rendering`).

README.md
jyasskin: Oooh, "clicks a link" is just wrong here. I meant to write about the page loading a portal and learning whether it's activatable, as discussed in #191 and #185. However, the fact that the initial load is done without credentials blocks this entirely, instead of just being "no worse than other mechanisms in the target platform". So ... I think I'll move this and the timing channel to the "blocked" list.

README.md
```diff
@@ -107,16 +107,35 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
...
 
+The Portals design is intended to comply with the [W3C Target Privacy Threat Model](https://w3cping.github.io/privacy-threat-model/). This section discusses the aspects of that threat model that are particularly relevant to Portals and how the design satisfies them.
+
+A portal can contain either a same-site or cross-site resource. Same-site portals don't present any privacy risks, but cross-site resources risk enabling [cross-site recognition](https://w3cping.github.io/privacy-threat-model/#model-cross-site-recognition) by creating a messaging channel across otherwise-partitioned domains. For simplicity, when a cross-site channel needs to be blocked, we also block it for same-site cross-origin portals.
+
+This design assumes that storage will be [partitioned](https://github.com/privacycg/storage-partitioning), even though that is not the status quo in all browsers. Because portaled documents can be activated into a top-level browsing context, they (eventually) live in the first-party storage partition of their origin. This means we have to prevent communication with the surrounding document to the same extent we prevent it with a cross-site link opened in a new tab.
```
jakearchibald, line 114: Are we going to follow the lead of partitioning here, or might we come up with a simpler system like "no storage until activation" rather than partition?

README.md
```diff
@@ -107,16 +107,35 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
...
+
+#### Channels that are blocked
+
+* [`postMessage()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) isn't allowed between a cross-origin portal and its container.
```
jakearchibald, line 118: The use of "cross-origin" in this line makes it sound like it _might_ be an exception to this from above:

> For simplicity, when a cross-site channel needs to be blocked, we also block it for same-site cross-origin portals.

README.md
```diff
@@ -107,16 +107,35 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
...
+  * The site creates a portal, and the target site decides between a 204 and a real response based on the user's ID. Or the site delays the response by an amount of time that depends on the user's ID. Because the portal load is done without credentials, the site can't get its user ID in order to make this sort of decision.
+* Side channels, further discussed in [Rendering](#rendering):
+  * Portal Resize: Javascript outside the portal could use a sequence of portal resizes to send a user ID, so a portal's content cannot observe any resizes after creation. If a portal is resized, that just rescales the view of the portal. For simplicity, this is the case for same-origin portals too.
```
jakearchibald, line 123: nit

```suggestion
  * Portal Resize: JavaScript outside the portal could use a sequence of portal resizes to send a user ID, so a portal's content cannot observe any resizes after creation. If a portal is resized, that just rescales the view of the portal. For simplicity, this is the case for same-origin portals too.
```

README.md
```diff
@@ -107,16 +107,35 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
...
+  * Portal initial size: Javascript outside the portal could use the initial size to send part of a user ID, so portals are always sized the same as the top-level tab that they'll be activated into and then scaled into the portal element. For simplicity, this is the case for same-origin portals too.
```
jakearchibald, line 124: nit

```suggestion
  * Portal initial size: JavaScript outside the portal could use the initial size to send part of a user ID, so portals are always sized the same as the top-level tab that they'll be activated into and then scaled into the portal element. For simplicity, this is the case for same-origin portals too.
```

README.md
jyasskin: I meant this to be compatible with "no storage until activation", as the partitioning explainer anticipates that some storage will be blocked from a third-party context: https://github.com/privacycg/storage-partitioning/issues/8. The "(eventually)" hints at this, but I wonder if you have an idea for being clearer about it, without using lots of words?

README.md
jyasskin: This is an _example_ of the above: it only truly needs to be blocked for cross-site portals, but we're also blocking it for same-site cross-origin portals.

README.md
```diff
@@ -107,16 +107,35 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
...
+
+#### Channels that are blocked
+
+* [`postMessage()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) isn't allowed between a cross-origin portal and its container.
+* Blocked by preventing fetches within portals from using credentials until they're activated:
+  * The sequence of portal loads: The source page can encode its user ID into the order in which a sequence of URLs are loaded into portals. To prevent the target from correlating this ID with its own user ID without a navigation, documents loaded into cross-origin portals are fetched without credentials and don't have access to either first-party storage or multi-keyed storage. They have to use [`requestStorageAccess()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/requestStorageAccess) (or another TBD mechanism) to get access while they're being activated.
```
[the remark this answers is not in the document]
jyasskin: Done.

README.md
jakearchibald: Not sure I can do fewer words, but maybe:

> This design assumes that storage will be partitioned/blocked in line with the browser's implementation (or lack thereof) of [storage-partitioning](https://github.com/privacycg/storage-partitioning).

README.md
jakearchibald: Maybe just:

* [`postMessage()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) isn't allowed between a portal and its container.

…removing "cross-origin", because it's implied above?

jyasskin: There are some restrictions below that that apply to same-origin portals, and I wanted to be clear which was which. I feel like "isn't allowed between a portal and its container" reads like it applies to same-origin portals too, which isn't the case.

README.md
jyasskin: I've added "or blocked" to try to capture this. How's the new text?

domenic:
Let's merge this :). Followup tweaks always welcome!

PR #210, opened 2020-06-16
Add description of accessibility defaults

a4sriniv:
Adds description of implemented default accessibility behaviour (default role and label) for portals.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,9 +159,14 @@ TODO:
 
 ### Accessibility
 
+From an accessibility perspective, portals behave like a single activatable element (similar to a button). As discussed in the section above, portals are not interactive and don't receive input events and focus. As a result, the embedded contents of a portal are not exposed as elements in the accessibility tree.
+
+Portals come with accessibility defaults right out of the box. Their default ARIA role is `"button"`, and they are therefore visible to screen-readers as a button by default. The portal element is tab-focusable and keyboard activatable by default (any registered `click` event handlers will be run when a user hits `<space>` or `<enter>` while focused on the element).
+
+Portals also compute a default label from their embedded contents (by either using the title of the embedded page or concatenating all the visible text in the portal's viewport if the page doesn't have a title). This label can be overridden by authors using the `aria-label` attribute.
+
 TODO:
 
-- Discuss role and description.
 - Discuss best practices for authors of portaling pages.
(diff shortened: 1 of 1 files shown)


yoavweiss:
@a4sriniv - Can you join the WICG to appease the IPR bots?

a4sriniv:
@yoavweiss I just created an account and linked it with GitHub, but I don't have permissions to get the bot to revalidate.

README.md
```diff
@@ -159,9 +159,14 @@ TODO:
 
 ### Accessibility
 
+From an accessibility perspective, portals behave like a single activatable element (similar to a button). As discussed in the section above, portals are not interactive and don't receive input events and focus. As a result, the embedded contents of a portal are not exposed as elements in the accessibility tree.
+
+Portals come with accessibility defaults right out of the box. Their default ARIA role is `"button"`, and they are therefore visible to screen-readers as a button by default. The portal element is tab-focusable and keyboard activatable by default (any registered `click` event handlers will be run when a user hits `<space>` or `<enter>` while focused on the element).
```
domenic, line 164: Since focusability, especially tab-focusability, is different per browser, we shouldn't make blanket statements here. The same goes for Space or Enter, I think. So I'd suggest stating something like "The portal element is intended by focusable and keyboard activatable in the same way as a button", or something like that?

yoavweiss:
@a4sriniv - did you join https://www.w3.org/community/wicg/participants ? The bots are still unhappy

a4sriniv:
@yoavweiss - ah I did not, I just sent a request to join.

README.md
a4sriniv: Updated. I was trying to indicate that it respects the requirements for the button role here: https://www.w3.org/TR/wai-aria-practices-1.1/#button, but I think your suggestion gets that across too.

README.md
```diff
@@ -159,9 +159,14 @@ TODO:
 
 ### Accessibility
 
+From an accessibility perspective, portals behave like a single activatable element (similar to a button). As discussed in the section above, portals are not interactive and don't receive input events and focus. As a result, the embedded contents of a portal are not exposed as elements in the accessibility tree.
```
domenic, line 162:
```suggestion
From an accessibility perspective, portals behave like a single activatable element (similar to a button). As discussed in the section above, the contents of portals are not interactive and don't receive input events and focus. As a result, the embedded contents of a portal are not exposed as elements in the accessibility tree.
```

Issue #174, opened 2020-01-24
Default click behaviour of portals


jakearchibald:
It feels like we have an opportunity to make these things useful even without JavaScript.

We could make their default click behaviour to activate. This means if the user clicks on a portal, and the site's JavaScript fails to call activate (eg, it throws an error, fails to load, or hasn't loaded yet), they still get to the site they clicked on.

This makes them behave kinda like links by default.

---

This also gives us an opportunity to provide defaults to shift+click, middle click etc. The complicating factor here is: What happens to a portal if it's opened in a new window?

I guess starting a fresh navigation in a new window is least destructive, but doesn't make use of the pre-loaded page.

Alternatively, we could activate the portal in the new window, and do 'something' with the portal in the page. I'm just not sure what.

a4sriniv:
We decided recently to make portals behave more like buttons instead. There are usually DOM changes/animations that need to be carried out before activation which would be skipped if we directly called activate on click.

We're making portals keyboard focusable by default, and making them simulate a click event/run click event handlers when they receive enter/space key events by default.

jakearchibald:
> There are usually DOM changes/animations that need to be carried out before activation which would be skipped if we directly called activate on click.

"Need to" or "should"? Is 'nothing happens' a better failure case than 'portal activates'?

a4sriniv:
In some scenarios, nothing happens is a better failure case (for example, some pages may want to limit activatation to some other gestures (like scrolling)). We could also have portals that authors never want to be activated.

jakearchibald:
Right, but `preventDefault()` can be used to stop the default behaviour (just like they are with links). The question is about which is the better failure case.

a4sriniv:
Hmm would `preventDefault()` work if the site's javascript throws an error/fails to load (going back to the inital premise here)?

I worry that authors may expect that they control exactly when a portal activates for the first time by default. Also I think if portals exist to enable seamless navigation (which differentiates them from links), we should have defaults that encourage authors to think about the transition, and not give a default behaviour that is abrupt.

But I think that the link behaviour is feasible, because we do intend to have browser triggered activations for back/forward navigations.

jakearchibald:
> Hmm would `preventDefault()` work if the site's javascript throws an error/fails to load (going back to the inital premise here)?

Only if it's called before the JS throws.

> Also I think if portals exist to enable seamless navigation (which differentiates them from links), we should have defaults that encourage authors to think about the transition, and not give a default behaviour that is abrupt.

[...]

PR #213, opened 2020-06-17
Add short description of accessibility best practices

a4sriniv:


README.md
```diff
@@ -184,9 +184,10 @@ Portals come with accessibility defaults right out of the box. Their default ARI
 
 Portals also compute a default label from their embedded contents (by either using the title of the embedded page or concatenating all the visible text in the portal's viewport if the page doesn't have a title). This label can be overridden by authors using the `aria-label` attribute.
 
+The current defaults should ensure that a portal can be accessed by screen readers and have a description without any work from authors. Authors should however implement a click event handler that activates the portal, to ensure they can be activated with screen readers. Portals that are hidden by their embedder until activation time should be marked with `aria-hidden="true"` by authors, so that they override the default behaviour and are invisible to screen readers as well.
```
domenic, line 187: > to ensure they can be activated with screen readers

Well, to ensure they can be activated by anybody, right? Might be worth linking to #174 here.

> Portals that are hidden by their embedder until activation time should be marked with `aria-hidden="true"` by authors

I'd generalize this to say that "should be hidden by authors to avoid interaction (using e.g. the `hidden=""` attribute, or `display: none`)" and then maybe mention that this will also hide them from the accessibility tree. In general, we want accessibility to be more automatic, and not require special `aria-*` attributes to get good behavior. That helps authors more easily create good experiences for all users. I think portals has a good story here, so we shouldn't imply that you need special `aria-hidden=""` management for them.

a4sriniv: > > to ensure they can be activated with screen readers
> 
> Well, to ensure they can be activated by anybody, right? Might be worth linking to #174 here.
>
Technically yes, but I intended this more for use cases where activation normally happens through some user gesture that isn't easy to reproduce with a screen reader (or isn't intuitive to do when using one). In particular I'm thinking about scroll/swipe triggered activations.

This has got me thinking more about #174 though, and I think that it's a great default from an accessibility point of view.

> > Portals that are hidden by their embedder until activation time should be marked with `aria-hidden="true"` by authors
> 
> I'd generalize this to say that "should be hidden by authors to avoid interaction (using e.g. the `hidden=""` attribute, or `display: none`)" and then maybe mention that this will also hide them from the accessibility tree. In general, we want accessibility to be more automatic, and not require special `aria-*` attributes to get good behavior. That helps authors more easily create good experiences for all users. I think portals has a good story here, so we shouldn't imply that you need special `aria-hidden=""` management for them.

That makes sense, updated.

domenic:
I pushed some minor tweaks for copy-editing and to expand on the points you made about user gestures. Yay, merging!

a4sriniv:
> I pushed some minor tweaks for copy-editing and to expand on the points you made about user gestures. Yay, merging!

Thanks 😄

PR #223, opened 2020-07-04
Define the activation behavior of a <portal> element.

### LINKED ISSUE #174: Default click behaviour of portals

jeremyroman:
Resolves #174.

***
<a href="https://pr-preview.s3.amazonaws.com/jeremyroman/portals/pull/223.html" title="Last updated on Jul 7, 2020, 5:21 PM UTC (533c846)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/223/7e745af...jeremyroman:533c846.html" title="Last updated on Jul 7, 2020, 5:21 PM UTC (533c846)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -630,6 +630,9 @@ spec:url; type:dfn; text:scheme
     portals-cross-origin-load.sub.html
   </wpt>
 
+  A <{portal}> element's [=activation behavior=] is to run the steps for
+  {{activate(options)}} with default options, discarding any thrown exception.
+
   The following events are dispatched on {{HTMLPortalElement}} objects:
 
   <table class="data" dfn-for="HTMLPortalElement">
--- a/index.bs
+++ b/index.bs
@@ -135,8 +135,8 @@ spec:url; type:dfn; text:scheme
 
   <section algorithm="portal-browsing-context-activate">
     To <dfn>activate a portal browsing context</dfn> |successorBrowsingContext| in
-    place of |predecessorBrowsingContext| with [=origin=] |sourceOrigin|, data
-    |serializeWithTransferResult|, and promise |promise|, run the following steps [=in parallel=]:
+    place of |predecessorBrowsingContext| with [=origin=] |sourceOrigin|, optional data
+    |serializeWithTransferResult|, and optional promise |promise|, run the following steps [=in parallel=]:
 
     1. [=Assert=]: The [=portal state=] of |predecessorBrowsingContext| is "`none`".
 
@@ -167,8 +167,8 @@ spec:url; type:dfn; text:scheme
 
         1. Let |dataClone| be null.
 
-        1. If |successorBrowsingContext|'s [=active document=]'s [=Document/origin=] is
-            [=same origin=] with |sourceOrigin|, then:
+        1. If |serializeWithTransferResult| is given and |successorBrowsingContext|'s [=active document=]'s
+            [=Document/origin=] is [=same origin=] with |sourceOrigin|, then:
 
             1. Let |deserializeRecord| be
                 [$StructuredDeserializeWithTransfer$](|serializeWithTransferResult|, |targetRealm|),
@@ -186,7 +186,7 @@ spec:url; type:dfn; text:scheme
 
         1. Set |event|'s [=PortalActivateEvent/successor window=] to |successorWindow|.
 
-        1. Set |event|'s [=PortalActivateEvent/activation promise=] to |promise|.
+        1. Set |event|'s [=PortalActivateEvent/activation promise=] to |promise|, if it is given, and null otherwise.
 
         1. [=Dispatch=] |event| to |successorWindow|.
 
@@ -217,7 +217,7 @@ spec:url; type:dfn; text:scheme
 
         1. Otherwise:
 
-            1. [=Queue a global task=] on the [=portal task source=] given
+            1. If |promise| is given, [=queue a global task=] on the [=portal task source=] given
                 |predecessorBrowsingContext|'s [=browsing context/active window=] to resolve
                 |promise| with undefined.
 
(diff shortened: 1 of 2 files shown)


domenic:
I think we need a bit more than this:

- Prefer not referencing public JS APIs, since they can be overwritten. Instead we should factor out a shared algorithm.
- We need to update the explainer. In particular:
  - https://github.com/WICG/portals#example references #174 
  - https://github.com/WICG/portals#interactivity should get updates and remove a TODO
  - https://github.com/WICG/portals#accessibility : we should consider making portals expose themselves as links now, instead of as buttons

jeremyroman:
> I think we need a bit more than this:
> 
> * Prefer not referencing public JS APIs, since they can be overwritten. Instead we should factor out a shared algorithm.

Hmm, I didn't think that was so. It refers to the steps associated with a WebIDL operation, not an ECMAScript function (and doesn't use [[Get]] or anything that would retrieve an author-overridden value).

If I were to factor it out, would throwing exceptions and creating promises still be appropriate in that algorithm, or would you want the error handling abstracted too? It'd be unfortunate if this simple change complicated stuff that much.

> * We need to update the explainer. In particular:
>   
>   * https://github.com/WICG/portals#example references #174
>   * https://github.com/WICG/portals#interactivity should get updates and remove a TODO
>   * https://github.com/WICG/portals#accessibility : we should consider making portals expose themselves as links now, instead of as buttons

It's not obvious to me whether/why this should be so. It seems separate from #174 to me.

jeremyroman:
also @a4sriniv do you have an opinion on button/link?

domenic:
> If I were to factor it out, would throwing exceptions and creating promises still be appropriate in that algorithm, or would you want the error handling abstracted too? It'd be unfortunate if this simple change complicated stuff that much.

Actually I think this indicates further how it'd be good to factor it out. As-is, it's not clear what happens if an exception is thrown, or what is done with the returned promise (especially if it's rejected; do we mark it as handled?). It'd be ideal to have an algorithm that was specific to user-initiated activations, which could, e.g., no-op for step 2, and assert instead of throw for steps 4 and 5.

> It's not obvious to me whether/why this should be so. It seems separate from #174 to me.

If you're referring to the button/link distinction only, then I agree we could do that separately. The other explainer updates should be done atomically with this, I think.

jeremyroman:
> > If I were to factor it out, would throwing exceptions and creating promises still be appropriate in that algorithm, or would you want the error handling abstracted too? It'd be unfortunate if this simple change complicated stuff that much.
> 
> Actually I think this indicates further how it'd be good to factor it out. As-is, it's not clear what happens if an exception is thrown, or what is done with the returned promise (especially if it's rejected; do we mark it as handled?). It'd be ideal to have an algorithm that was specific to user-initiated activations, which could, e.g., no-op for step 2, and assert instead of throw for steps 4 and 5.

The bulk of the work is already in the **activate a portal browsing context** algorithm. I can duplicate the `activate(options)` steps; I don't see a natural way to factor them out (without reference to DOMException etc).

I will note that the assertion you mention isn't valid, because activation behavior can still be triggered (e.g. by calling `click()` from script) on elements which are not in a browsing context. I do think that in practice UAs will end up printing more or less the same message as a console warning/error that they would have attached to an exception message, but I concede that fact doesn't need to be in the spec.

Let me see what that looks like.

> > It's not obvious to me whether/why this should be so. It seems separate from #174 to me.
> 
> If you're referring to the button/link distinction only, then I agree we could do that separately. The other explainer updates should be done atomically with this, I think.

Yes, sorry, that is what I meant.

a4sriniv:
> also @a4sriniv do you have an opinion on button/link?

The default activation does satisfy the "navigate on click" requirement for links, but we still can't activate portals into a new tab/window (to mirror opening links in a new tab), and I'm not sure how important that is for links.

index.bs
```diff
@@ -630,6 +630,33 @@ spec:url; type:dfn; text:scheme
...
+
+    1. Let |sourceOrigin| be [=this=]'s [=relevant settings object=]'s
+        [=environment settings object/origin=].
+
+    1. Run the steps to [=activate a portal browsing context|activate=] |portalBrowsingContext|
+        in place of |predecessorBrowsingContext| with |sourceOrigin|.
+
+    <div class="note">
+      This is substantially similar to the steps in {{activate(options)}}, with default options.
+      User agents may wish to display suitable console messages under the same conditions that would
```
domenic, line 655:
```suggestion
      User agents might wish to display suitable console messages under the same conditions that would
```

Ideally don't use may/must/should in notes.

index.bs
```diff
@@ -630,6 +630,33 @@ spec:url; type:dfn; text:scheme
     portals-cross-origin-load.sub.html
   </wpt>
 
+  <section algorithm="htmlportalelement-activation-behavior">
+    A <{portal}> element's [=activation behavior=] is to run the following steps:
+
+    1. Let |portalBrowsingContext| be the [=guest browsing context=] of [=this=].
+
+    1. If |portalBrowsingContext| is null, abort these steps.
```
domenic, line 638: Style nit these days:

```suggestion
    1. If |portalBrowsingContext| is null, return.
```

index.bs
```diff
@@ -630,6 +630,33 @@ spec:url; type:dfn; text:scheme
...
+
+    1. Let |predecessorBrowsingContext| be the [=document browsing context|browsing context=] of
+        [=this=]'s [=node document=].
+
+    1. If |predecessorBrowsingContext| is null, abort these steps.
```
domenic, line 643:
```suggestion
    1. If |predecessorBrowsingContext| is null, return.
```

index.bs
```diff
@@ -630,6 +630,33 @@ spec:url; type:dfn; text:scheme
...
+
+    1. If the [=portal state=] of |predecessorBrowsingContext| is not "`none`", abort these steps.
```
domenic, line 645:
```suggestion
    1. If the [=portal state=] of |predecessorBrowsingContext| is not "`none`", return.
```

index.bs
```diff
@@ -630,6 +630,33 @@ spec:url; type:dfn; text:scheme
     portals-cross-origin-load.sub.html
   </wpt>
 
+  <section algorithm="htmlportalelement-activation-behavior">
+    A <{portal}> element's [=activation behavior=] is to run the following steps:
+
+    1. Let |portalBrowsingContext| be the [=guest browsing context=] of [=this=].
```
domenic, line 636: We're not in a Web IDL method, so `[=this=]` doesn't exist. Probably it's best to say "A `portal` element _el_'s activation behavior is to run the following steps:" and then use _el_ throughout.

index.bs
jeremyroman: Sure, okay. I don't think there's much risk of it being misinterpreted as normative, but "might wish" works fine too. :)

index.bs
jeremyroman: Done.

jeremyroman:
PTAL

domenic:
Spec LGTM. You up for the explainer updates too?

jeremyroman:
Man, I thought I'd done that but I apparently only imagined it. Done.

PR #231, opened 2020-07-22
Expand on restrictions to portaled content

### LINKED ISSUE #211: Portals should not have influence outside their rendered area
My assumption is the portal will not be able to do the following:

- `<a target="top">`: Navigates portal.
- `window.open`: Returns null.
- `el.requestFullscreen`: Rejects.
- `PresentationRequest`: Various methods reject.

Any others I'm missing?

I don't know if playing sound is something we should prevent, or mute. It isn't clear to me what should happen if a page is activated and then reportaled.

domenic:
- Expands the "Privacy threat model and restrictions" section with storage blocking and more detail on the attacks
- Adds "Other restrictions while portaled" which covers some of the TODOs from the old section "Permissions and policies", and closes #211.
- Introduces a placeholder "Opt-in" section since we need to refer to that a few times.
- Expands the "Summary of differences between portals and iframes" section to talk about these.

---

Reviewers: there are two main things to focus on here.

- The usual more fine-grained review. Especially around the storage restrictions, I want to make sure I explained things in a way that makes sense.

- The design questions. I've put forth some concrete plans to solve #211 here, which leans heavily on Permissions Policy and the Permissions API to identify disallowed things. Is that the right approach?


README.md
```diff
@@ -107,46 +112,72 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
...
+  - [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API): Won't give visibility information to script inside the portal, to avoid occlusion being used to send information.
+  - [Page Visibility](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) and the timing of `requestAnimationFrame()` callbacks match the visibility of the top-level page, as in iframes, to prevent the page from encoding messages in visibility changes. However, this leads to the possibility that a portal and containing page could use the timing of user-caused visibility changes to join the user across site boundaries. Whether and how to prevent this is [still under discussion](https://github.com/WICG/portals/issues/193#issuecomment-639768218).
+
+#### Communications channels that match navigation
+
+As mentioned above, we prevent communications "to the same extent we prevent it with a cross-site link opened in a new tab". In particular:
```
jeremyroman, line 154: nit: quotation marks probably aren't necessary to quote earlier in the same document?

README.md
```diff
@@ -197,6 +228,16 @@ TODO:
...
 - Maybe compare/contrast with iframes. Is it helpful or confusing to talk about how portals are more like popups than iframes? (I.e. they are top-level browsing contexts in spec terms.)
 
+### Opt-in
+
+Because of the [restrictions](#other-restrictions-while-portaled) on portaled content, especially the [storage restrictions](#storage-access-blocking), most existing content is not prepared to be rendered in a portal. As such, content will need to opt-in to being portaled; any content that does not so opt-in will cause the portal to fail to load.
+
+TODO:
+
+- Actually describe the opt-in, once we decide what it is.
+- Explain relation to `X-Frame-Options` and CSP `frame-ancestors`. These opt-outs probably become unnecessary if we have an explicit opt-in?
```
jeremyroman, line 238: Very much TBD, IMO. Might content want to allow itself to be prerendered offscreen but not displayed in a <portal>? My instinct is that if the opt in is to the generic "restricted prerendering" mode then it doesn't subsume XFO/CSP. We could add an explicit opt-in to portal embedding to that, perhaps.

That said, I think (but haven't thought deeply) that without credentials you would think there's less risk than XFO/frame-ancestors have historically protected against, insofar as the rendering it shows you is likely no different than if the referrer had taken their own screenshot of the web page and embedded that. We could consider revisiting this. Maybe file an issue if you think we should reopen that?

domenic: I'd planned to treat it as part of the prerendering opt-in discussions, but yeah, we can have a dedicated issue.

README.md
```diff
@@ -107,46 +112,72 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
...
+
+For a more concrete example, consider `https://aggregator.example/` which wants to prerender this GitHub repository using a `<portal>`. To make this work, GitHub would need to add the opt-in to allow the page to be portaled. Additionally, GitHub should add code to adapt their UI to show the logged-in view upon activation, by removing the "Join GitHub today" banner, and retrieving the user's credentials from storage and using them to replace the signed-out header with the signed-in header. Without such adapter code, activating the portal would show the user a logged-out view of GitHub in the top-level tab that the portal has been activated into. This would be a bad and confusing user experience, since the user is logged in to GitHub in all of their other top-level tabs.
+
+#### Communications channels that are blocked
+
+- [`postMessage()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) isn't allowed between a cross-origin portal and its container.
+- Blocked by preventing fetches within portals from using credentials until they're activated:
```
bokand, line 143: Drive-by: I was reading the explainer today and found this section confusing so this review seems like an ideal time to ask questions and make clarifications.

> Blocked by preventing fetches within portals from using credentials until they're activated

This sentence is awkward. Does "Fetches within an unactivated portal are performed without credentials" describe what this is saying?

bokand: Also, are the two sub-bullets below examples of potential communication channels if a portal can perform credentialed requests? If so, perhaps:

"Fetches within an unactivated portal are performed without credentials. Credentialed fetches can be used for cross-site recognition, for example:"

README.md
```diff
@@ -107,46 +112,72 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
...
+  - The sequence of portal loads: The source page can encode its user ID into the order in which a sequence of URLs are loaded into portals. To prevent the target from correlating this ID with its own user ID without a navigation, a document loaded into a cross-origin portal is fetched without credentials and doesn't have access to either first-party storage or multi-keyed storage. It has to use [`requestStorageAccess()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/requestStorageAccess) (or another TBD mechanism) to get access when it's activated.
```
bokand, line 144: > The source page can encode its user ID into the order in which a sequence of URLs are loaded into portals

I don't understand this...is the plural "portals" suggesting that there are multiple `<portal>` elements and the order of their navigations can be used to signal a value to the target server? Or is it a sequence of navigations on a single `<portal>` to different URLs, the order of which is used to communicate the value.

Either case raises the question: why use a sequence of URLs rather than just putting the user id in the URL? e.g. 

`portal.src = 'https://tracker.com/user/bokand';

bokand: > To prevent the target from correlating this ID with its own user ID without a navigation

What navigation is  "without a navigation" referring to? Navigating to a different URL in the portal? Or activating the portal? Either way, don't we want to prevent correlating IDs across sites in both these cases?

README.md
```diff
@@ -107,46 +112,72 @@ The general idea of portals is summed up above: inset pre-rendering, activation,
...
+  - The site creates a portal, and the target site decides between a 204 and a real response based on the user's ID. Or the site delays the response by an amount of time that depends on the user's ID. Because the portal load is done without credentials, the site can't get its user ID in order to make this sort of decision.
```
bokand, line 145: IIUC, the first instance of "The site" in this point refers to the source site but the other two refer to the target site which is mildly confusing.

Also, "site" could be interpreted in the sense of eTLD+1, especially since it's inconsistent with the point above which uses "page" for the same concept.

Maybe change these instances to "target|source page|server"?

README.md
bokand: Also, also, presumably this is all moot for same-origin fetches - do we still fetch uncredentialed? What about the other restrictions below?

domenic: I'll clarify this in the manner you suggest; I agree these are good changes.

README.md
domenic: I think the following answers are correct, but I am not sure:

> I don't understand this...is the plural "portals" suggesting that there are multiple `<portal>` elements and the order of their navigations can be used to signal a value to the target server? Or is it a sequence of navigations on a single `<portal>` to different URLs, the order of which is used to communicate the value.

Either would work.

> Either case raises the question: why use a sequence of URLs rather than just putting the user id in the URL? e.g.

I don't _think_ this case ("link decoration") is explicitly addressed yet. @jyasskin, do you agree? We talk about link decoration in the activation case, where activating is similar to clicking a link. But just the pure embedding case, with no activation, doesn't seem to be called out explicitly.

> What navigation is "without a navigation" referring to? Navigating to a different URL in the portal? Or activating the portal?

Activating the portal.

Once you activate the portal, you end up in the "Communications channels that match navigation" case; activating a portal should leak no more information than clicking a link does. The idea is that both are highly user-visible actions.

README.md
domenic: Will do!

README.md
domenic: > I don't think this case ("link decoration") is explicitly addressed yet. @jyasskin, do you agree?

In particular, currently we have

> **The URL of the activated portal** and the referring URL are available to portals to the same extent they're available to normal navigations. Solutions to link decoration will apply to both.

but we don't talk about the URL of the portal pre-activation.

jeremyroman: Without activation the other site can't access its own credentials, which means this is more similar embedding a subresource such as an image (which in the future won't have first-party cookies). With activation it's the general link decoration problem (with work going on in other groups to address that).

domenic: Ah, right. With no activation ever, there's no threat. With eventual activation, then there's the threat of

> If we allowed communication, then the portaled content could be given the user ID from the host site. Then, after activation gives the portaled page access to first-party storage, it would join that user ID with information from its own first-party storage to perform cross-site tracking.

i.e. this opens us up to cross-site tracking in the same way that `<a>` does.

bokand: > I don't think this case ("link decoration") is explicitly addressed yet.

Ah, thanks - I'd seen a few references to "link decoration" but wasn't certain what that referred to :)

> i.e. this opens us up to cross-site tracking in the same way that <a> does.

Got it - clicking on the portal/link implies the user has some relationship/trust with the destination so that's less of a problem than pre-activation.

domenic: I spelled this out a bit more:

> Note that since a non-activated portal has no storage access, it cannot join any information stored in the URL with any of the portaled site's data. So it's only activation, which gives full first-party storage access, which creates a navigation-equivalent communications channel. This equivalence makes sense, as activating a portal is much like clicking a link.

domenic:
I'll go ahead and merge this, but I'd still love review from our permissions/policy/privacy experts, @jyasskin and @clelland, when they have time. Thanks to @bokand for the fresh set of eyes that led to some good improvements.

PR #233, opened 2020-07-22
Consistently refer to example.com in the explainer

kjmcnee:
The first example in the explainer uses www.example.com and
example.com interchangeably. We now just use example.com for
consistency, since these are technically different origins.


kjmcnee:
Adding @jeremyroman . PTAL.

This is just a trivial change to make sure I can contribute to the repro without issue before submitting meaningful changes.

Issue #232, opened 2020-07-22
Reconsider X-F-O and CSP frame-ancestors restrictions in light of opt-in


domenic:
As discussed internally a good amount, and surfaced into the explainer in https://github.com/WICG/portals/pull/231, portaled content will need to opt-in to being portaled.

I believe this should allow us to remove the need to support `X-Frame-Options` and CSP's `frame-ancestors` opt-outs, as portals no longer becomes a way to embed content that might not want to be embeddable. This allows those two controls to stick to controlling frames, and not awkwardly also control portals, which was already a mismatch (e.g. there are definitely sites that don't want to be framed but do want to be portaled; maybe also vice-versa).

jakearchibald:
There's a little mismatch here. `X-Frame-Options: DENY` is the most popular option, due to the legacy issues with `SAMEORIGIN`. Will that also block same-origin portals? Are we requiring an opt-in for same-origin portals?

jeremyroman:
I think `X-Frame-Options: DENY` does block same-origin portal embedding, and semantically that kinda is what they're asking for. I don't know what these legacy issues are, but couldn't authors change to `X-Frame-Options: SAMEORIGIN` or `Content-Security-Policy: frame-ancestors 'self'`?

jakearchibald:
Older browsers didn't check check all ancestors were same origin when using `X-Frame-Options: SAMEORIGIN`. I'm not sure if it's still a problem today (IE, maybe?).

Issue #228, opened 2020-07-10
activate() should fulfill even if the inner document load failed due to DNS/x-frame-options etc etc


jakearchibald:
Only just occurred to me that iframes fire their load event even in the case of navigation 'failures'. If we reject in similar cases we're revealing something new.

domenic:
This is somewhat contentious and non-interoperable, actually; see https://github.com/whatwg/html/issues/125, https://github.com/whatwg/html/issues/1230, and https://github.com/whatwg/html/pull/4247.

jakearchibald:
Hah, perfect. I guess we should just do whatever the spec does right now, and change whenever the spec changes.

jeremyroman:
It will certainly be observable to them that they weren't unloaded. Unlike the `load` event, `activate` does commit them to going there if it did succeed in loading and so is less useful for passively observing XFO etc than `<iframe onload>`.

PR #234, opened 2020-07-23
Provide initial explanation of session history and navigation

kjmcnee:
The more complicated scenarios still need explanation, but this conveys the overall design.


kjmcnee:
Adding @jakearchibald . PTAL.

README.md
```diff
@@ -189,13 +189,24 @@ Authors should respect the `prefers-reduced-motion` media query by conditionally
...
+
+Navigations within portals are subject to certain security restrictions. Notably, content that does not allow embedding (e.g. via `X-Frame-Options` [see above](#permissions-and-policies)) cannot load in a portal. Furthermore, both the portal host and the portal content must use an HTTP(S) scheme. This is in order to establish the origin of the content for it to be used safely.
```
jakearchibald, line 198: Do we want to mention that we might introduce a way for content to opt into portal use despite `X-Frame-Options`?

> This is in order to establish the origin of the content for it to be used safely.

Struggling to parse this line.

README.md
```diff
@@ -189,13 +189,24 @@ Authors should respect the `prefers-reduced-motion` media query by conditionally
...
+
+Navigation errors within portals may cause portal activation to be rejected. Instead of, for example, the user agent showing an error page to the user as with a conventional navigation, the promise returned by the activate method allows a page to gracefully handle the rejection. Furthermore, user agents have existing limitations on navigations initiated by the page where they may be ignored if they are considered to conflict with a user's intent to perform a different navigation. Such cases are not described by the existing navigation spec (see [#218](https://github.com/WICG/portals/issues/218)), but portal activations are subject to these limitations. In the case where another navigation takes precedence over portal activation, the promise returned by the activate method rejects.
```
jakearchibald, line 200: > Navigation errors within portals may cause portal activation to be rejected.

I think that may be exposing something we don't currently expose.

README.md
domenic: I think it might be best to omit this paragraph entirely, given that discussions are still ongoing in https://github.com/WICG/portals/issues/232

README.md
```diff
@@ -189,11 +189,11 @@ Authors should respect the `prefers-reduced-motion` media query by conditionally
...
...d take them back to the last thing they saw. Unfortunately, as mentioned [below](#summary-of-differences-between-portals-and-iframes), the mental model of a portal may differ between users and web developers. While developers will see portals as separate contexts navigating independently from their host context, users will not. This means that the portal context should not have an independent session history which we aggregate with its host context to present to the user. To illustrate how this could cause problems, consider `example.com/a.html` loading `b.html` in a portal, then navigating the portal to `c.html`, then activating the portal. If the user pressed the back button, it would not be appropriate to load `b.html`, since (1) this makes for bad UX as this is a state the user has not seen before and, (2) this is a concern for privacy as the user's browsing history would contain a page they did not intend to visit. Even for states that the user has seen previously, having a single navigation reintroduce potentially multiple history states maps poorly to the web's chronological history model, as opposed to the mental models for tab switching or mobile device style multitasking.
```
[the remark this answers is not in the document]
domenic: +1. I found the discussion of both web developer and user issues in this paragraph a bit confusing. Jake's suggestion focuses it entirely on the user-facing issues which is great.

Another suggestion: after

> To illustrate how this could cause problems, consider `example.com/a.html` loading `b.html` in a portal, then navigating the portal to `c.html`, then activating the portal. If the user pressed the back button, it would not be appropriate to navigate the top level page to `b.html`, since this is a state the user has not seen before.

which says what the proposed model will _not_ do, add a sentence stating what it _will_ do.

README.md
domenic: Worth filing an issue and pointing to it from here, perhaps.

kjmcnee: The same information would be exposed in other ways. For example, |fetch('https://doesnotexist.example.com/')| rejects.

domenic: But you can't tell if that fetch fails because of lack of CORS headers, or because of an actual network error.

jakearchibald: I think there are cases where that isn't true. `X-frame-options: DENY` wouldn't cause fetch to fail, but it would cause a portal to fail, for instance.

kjmcnee: Oops, I missed issue #228 which talks about this. I'll reference it here for now.

README.md
```diff
@@ -189,13 +189,25 @@ Authors should respect the `prefers-reduced-motion` media query by conditionally
...
+
+We want the history model for portals to conform to users' expectations of the back button. Specifically, the back button should take them back to the last thing they saw. To accommodate this, the portal context should not have an independent session history which we aggregate with its host context to present to the user. To illustrate how this could cause problems, consider `example.com/a.html` loading `b.html` in a portal, then navigating the portal to `c.html`, then activating the portal. If the user pressed the back button, it would not be appropriate to navigate the top level page to `b.html`, since this is a state the user has not seen before. Even for states that the user has seen previously, having a single navigation reintroduce potentially multiple history states maps poorly to the web's chronological history model, as opposed to the mental models for tab switching or mobile device style multitasking.
```
domenic, line 196:
```suggestion
We want the history model for portals to conform to users' expectations of the back button. Specifically, the back button should take them back to the last thing they saw. To accommodate this, the portal context should not have an independent session history which we aggregate with its host context to present to the user. To illustrate how this could cause problems, consider `example.com/a.html` loading `b.html` in a portal, then navigating the portal to `c.html`, then activating the portal. If the user pressed the back button, it would not be appropriate to navigate the top level page to `b.html`, since this is a state the user has not seen before. Instead, they navigate the top-level page back to `a.html`, with either `b.html` in the portal (if the back-navigation is a fresh load) or `c.html` in the portal (if the back-navigation uses bfcache). Even for other examples involving states that the user has seen previously, having a single navigation reintroduce potentially multiple history states would map poorly to the web's chronological history model, as opposed to the mental models for tab switching or mobile device style multitasking.
```

Is this right?

README.md
kjmcnee: Removed the specific mention of XFO and now just reference the relevant section.

Reworded the bit about origins. If we need explain why origins are important here, we could reference https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/GbVcuwg_QjM%5B1-25%5D , otherwise I can leave it as is.

domenic: That's a good link, but I think it's OK to omit for now.

README.md
kjmcnee: Essentially yes. But the adoption/bfcache part is complicated enough that, IMO, it's worth describing separately from the point being made here.

Reworded this a bit.

kjmcnee:
> Looks great. Let's give folks a chance to look at this over the weekend/Monday morning, then merge.

I'll be OOO until August 4. Feel free to merge if it looks okay while I'm away.

README.md
```diff
@@ -189,13 +189,25 @@ Authors should respect the `prefers-reduced-motion` media query by conditionally
...
+
+Navigation errors within portals may cause portal activation to be rejected. Instead of, for example, the user agent showing an error page to the user as with a conventional navigation, the promise returned by the activate method allows a page to gracefully handle the rejection. Furthermore, user agents have existing limitations on navigations initiated by the page where they may be ignored if they are considered to conflict with a user's intent to perform a different navigation. Such cases are not described by the existing navigation spec (see [#218](https://github.com/WICG/portals/issues/218)), but portal activations are subject to these limitations. In the case where another navigation takes precedence over portal activation, the promise returned by the activate method rejects.
+
 TODO:
 
-- One-sentence reiteration of how activation is a navigation.
-- Talk about how session history is tricky.
 - Talk about how bfcache is tricky.
+- Describe reversing transitions on back with bfcache and adopted portals.
```
jeremyroman, line 205: It's looking less and less likely that readoption (re-adoption, not read-option) will be part of an MVP. If so this might be punted for awhile.

jakearchibald: fwiw I don't think adoption should be required for reverse transitions

PR #235, opened 2020-07-24
Explain relation to CSP, document policy, and sandboxing

domenic:
Original PR text:

This proposal is a result of me going through the CSP specs and seeing what made sense. Much if it hasn't been discussed before, so review on this PR should also include high-level critiques of the proposal. I just thought it'd be simpler to discuss a concrete PR instead of an issue.

Update:

I've morphed this PR to cover CSP, document policy, and sandboxing all together, as there are cross-cutting concerns.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ Note that since a non-activated portal has no storage access, it cannot join any
 
 Apart from the privacy-related restrictions to communications and storage, while portaled, pages are additionally restricted in various ways, to prevent them from interfering with content outside of their rendering area:
 
-- Portals are top-level browsing contexts, and given their own [browsing context group](https://html.spec.whatwg.org/multipage/browsers.html#groupings-of-browsing-contexts). This means they cannot synchronously access their embedder, e.g. to navigate it or modify its DOM (even when same-origin).
+- Portals are top-level browsing contexts, and given their own [browsing context group](https://html.spec.whatwg.org/multipage/browsers.html#groupings-of-browsing-contexts). This means they cannot synchronously access their embedder, e.g. to navigate it or modify its DOM.
 
 - Any features that are controlled by the [Permissions API](https://w3c.github.io/permissions/) ([list](https://w3c.github.io/permissions/#permission-registry)) will be automatically denied without prompting.
 
@@ -180,9 +180,7 @@ After activation, these restrictions are lifted: the portaled content is treated
 
 _Note: ideally Permissions Policy would become a superset of the Permissions API and iframe sandboxing. Then we could use that infrastructure as the single place to impose and lift these restrictions. Currently that is not the case, so the spec will be more messy._
 
-TODO:
-
-- Do all of these restrictions also apply to same-origin portals? I lean towards "yes"?
+All of these restrictions apply uniformly to both same-origin and cross-origin portals, for simplicity. We could consider lifting some of them for same-origin portals in the future, if use cases arise.
 
 ### Rendering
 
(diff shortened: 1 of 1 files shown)


domenic:
One choice I made was to not have [`prefetch-src`](https://w3c.github.io/webappsec-csp/#directive-prefetch-src) affect portals, since it seemed like using the separate `portal-src` category would give better control. Expanding `prefetch-src` from just `<link>`s to `<portal>`s seemed like a net negative. But, I'm open to other opinions.

domenic:
Also worth discussing is whether `frame-src` should affect portals. I propose that it does not; the split of `child-src` into `frame-src`, `object-src`, and `worker-src` makes me think that more granular categories are preferred, and so `portal-src` should be a sibling of `frame-src`/`object-src`/`worker-src` instead of a child of `frame-src`.

jeremyroman:
@lucasgadani was involved when we discussed this previously. I remember being of the view that we should probably respect existing `frame-src` for consistency e.g. with `frame-ancestors` (and I think we do), but I admit that in this case since we're talking about the content that's choosing to use `<portal>` it may make sense to split the directive. Certainly the way you propose is the best in the abstract, it's just a matter of whether this has adverse consequences for existing content.

jakearchibald:
There's also navigate-to will need to check portals when they activate, which means creating a new directive algorithm, since it would need to analyse the portal URL rather than a response object.

domenic:
Tagging in @a4sriniv for his specific review

README.md
```diff
@@ -252,6 +255,40 @@ TODO:
...
+
+To conclude, instead of giving embedders this control as iframes do, we believe that the browser can take the role of mitigating any problematic features. For example, instead of requiring embedders to use `sandbox=""` to turn off modal `alert()`/`confirm()`/`prompt()` dialogs, or permissions policy to turn off autoplaying media, those features are [always disabled](https://github.com/WICG/portals#other-restrictions-while-portaled) in pre-activation portals. And because portals are isolated from communicating with their embedder pre-activation, any problems which CSP Embedded Enforcement would attempt to protect against will instead be caught by this communications barrier and prevented from impacting the embedder.
```
a4sriniv, line 290: I think the behaviour for features controlled by permissions policy and permissions api is well defined (for existing features and for when we add any new features / change defaults), but I'm not super clear about portals do for the different `sandbox=` attributes. IIUC, we handle each attribute on a case by case basis, some make sense to enforce for portals, and some don't. If that's true, maybe it's worth adding a TODO to elaborate on the different sandbox attributes and which ones portals will be enforcing.

a4sriniv: Although, now that I actually look through all the possible sandbox values, I think `allow-forms` is the only behaviour that isn't really called out/explained in the explainer.

README.md
```diff
@@ -166,20 +168,21 @@ Note that since a non-activated portal has no storage access, it cannot join any
...
 
+_Note: ideally Permissions Policy would become a superset of the Permissions API and iframe sandboxing. Then we could use that infrastructure as the single place to impose and lift these restrictions. Currently that is not the case, so the spec will be more messy._
+
 TODO:
 
-- Do these restrictions also apply to same-origin portals?
-- Discuss CSP, document policies, `sandbox=""`, referrer policy... any other policies? Might not belong in this section.
-- Briefly discuss permissions and policies that people might want to impose on portals, including some that aren't proposed yet (like blocking scripts or network access). Note that these will be developed independently, for iframes and portals alike.
-- Ideally Permissions Policy would become a superset of the Permissions API, and we could use that infrastructure exclusively to shut things down. Currently that is not the case so we have to consult both.
+- Do all of these restrictions also apply to same-origin portals? I lean towards "yes"?
```
jeremyroman, line 185: I also lean towards "yes", though for same-origin it might be more reasonable to allow permission policy delegation in the future (insofar as the origin pair for the permission grant is the same etc, so keeping the permission across activation is sensical), subject to the discussions @a4sriniv had with permissions folk. But I'm not aware of that being a priority so punting seems cool.

domenic:
I'm going to file a tracking issue on the fact that the CSP design doesn't feel completely settled, link to it, and then merge this.

README.md
domenic: Right, I did a more full analysis in https://github.com/WICG/portals/issues/237. My conclusion was

> If we want to add controls for preventing form submission, plugin usage, scripting, document.domain, or autoplay/autofocus, we use document policy for that in the future.

although I realize now autoplay is controlled by permissions policy that case in particular is already prevented.

README.md
domenic: Let's make it so for now.

README.md
domenic: Hmm, although I guess the latest direction is to not allow document policy. Let's continue discussion in #237.

PR #236, opened 2020-07-24
Outline Document Policy integration

domenic:
This seems pretty straightforward, in the end.

Note to self: I've targeted this PR at the `csp` branch purely to avoid merge conflicts, and make the diff simpler. If this becomes ready to merge before #235, then we'll want to do some rebasing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Portals encompass some of the use cases that iframes currently do, but with bett
   - [Session history, navigation, and bfcache](#session-history-navigation-and-bfcache)
   - [Opt-in](#opt-in)
   - [CSP integration](#csp-integration)
+  - [Document Policy integration](#document-policy-integration)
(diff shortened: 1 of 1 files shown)


jeremyroman:
Doesn't seem unreasonable -- if we want to support document policy, I think this makes sense. I have some hesitation rooted in the unresolved questions about how pre-activation policies end up being integrated, and ecosystem questions about whether including this in an MVP will encourage required document policies in cases where the referrer could have permitted a broader swathe of content.

jeremyroman:
Maybe we should hold off on merging until we resolve those?

domenic:
Yeah, agreed, let's hold off until then.

domenic:
I'm going to move this into #235 as with recent changes I'll want to talk about CSP and document policy together, so two separate PRs is not managable.

PR #246, opened 2020-08-25
Specify window.close() to no-op in unactivated portals

### LINKED ISSUE #243: interactions with window.close
The spec should specify what happens when `window.close()` is invoked on a portal's window, or while activation is in progress. See https://crbug.com/1110926 for a Chromium bug that arose from this being ill-specified.

domenic:
Closes #243.

Note: this specifies that portals in the state "orphaned" are also not closable. I don't believe we have tests for that, but we should, IMO.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/portals/pull/246.html" title="Last updated on Aug 26, 2020, 6:17 PM UTC (49a8aef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/246/863124f...49a8aef.html" title="Last updated on Aug 26, 2020, 6:17 PM UTC (49a8aef)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -950,6 +950,24 @@ spec: ecma-262; urlPrefix: http://tc39.github.io/ecma262/
       </tr>
     </tbody>
   </table>
+
+  APIs for creating and navigating browsing contexts by name {#patch-window-apis}
+  -------------------------------------------------------------------------------
+
+  Modify the definition of <a spec=HTML>script-closable</a> to prevent portals from closing
+  themselves pre-activation:
+
+  A [=browsing context=] is <dfn noexport>script-closable</dfn> if either of the following is true:
+
+  * it is an [=auxiliary browsing context=] that was created by script (as opposed to by an action
+    of the user); or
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -950,6 +950,24 @@ spec: ecma-262; urlPrefix: http://tc39.github.io/ecma262/
       </tr>
     </tbody>
   </table>
+
+  APIs for creating and navigating browsing contexts by name {#patch-window-apis}
+  -------------------------------------------------------------------------------
+
+  Modify the definition of <a spec=HTML>script-closable</a> to prevent portals from closing
+  themselves pre-activation:
```
lucasgadani, line 958: Maybe we should word as closing when in a portal browsing context, as opposed to pre-activation? Mainly because a portal could be adopted, which may count as post-activation, but we would still want to prevent it from self-closing.

domenic: Yep, good call; I was trying to be informal in the informal intro, and I ended up being confusing.

PR #253, opened 2020-09-22
Delay activation on initial loading; src="" changes reset portal BC

### LINKED ISSUE #197: All src changes should reload the portal
```js
const portalURL = 'https://example.com/';
const portal = document.createElement('portal');
document.body.append(portal);
portal.src = portalURL + '#someData';
// Then, later:
portal.src = portalURL + '#someMoreData';
```

If changing the hash part of the URL _does not_ cause a reload, then it creates a data channel between the page and the portal.

We should ensure that all src changes reload the portal.

### LINKED ISSUE #227: What happens if activate is called while the portal is navigating
We decided in https://github.com/WICG/portals/issues/191 that the timing of activating a portal shouldn't be down to guesswork.

```js
portal.src = otherURL;
portal.activate();
```

What should happen here? Should it wait for `otherURL` to fetch, or should it try and activate the current context (which may be absent)?

Similarly, what should we do if the portal is undergoing a navigation that was triggered internally?

---

If the src is set, it feels like activation should wait for that action to settle.

If the portal is navigating by some other means, waiting seems like an information leak. The alternative would be to abort that navigation, and activate the current document. This is giving the host page some control over an action stated within the portal, but it seems `window.stop()` can already do that? Thoughts @jyasskin?

domenic:
Closes #197 by causing all changes to `src=""` to close the portal browsing context, and re-create it.

Closes #227 by causing `activate()` to delay promise resolution/rejection until the initial load in a new browsing context has finished.

<details>
<summary>Old content</summary>
@jakearchibald, @kjmcnee, please take a look. Other reviewers also appreciated.

Two main things to consider:

- Is this the behavior we want? I know @jakearchibald was skeptical about the "After the initial load" behavior, but I think it's better, and I believe @kjmcnee indicated that it's implementable. The alternative is to delay activation like we do with initial load, but that seems like a worse user experience.

- Is this explainer structure reasonable? I'm wondering if, e.g., we should move all discussion of activation failure from the bfcache section into here, or merge this with the bfcache section somehow. Also the flow feels a bit weird, in that we have to spend so much time up front explaining closed portals. But I couldn't find a better place to put it.
</details>

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/portals/pull/253.html" title="Last updated on Sep 28, 2020, 8:46 PM UTC (9c244a8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/253/703914f...9c244a8.html" title="Last updated on Sep 28, 2020, 8:46 PM UTC (9c244a8)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,6 +300,67 @@ An additional reason for avoiding these mechanisms is that it makes writing port
 
 To conclude, instead of giving embedders this control as iframes do, we believe that the browser can take the role of mitigating any problematic features. For example, instead of requiring embedders to use `sandbox=""` to turn off modal `alert()`/`confirm()`/`prompt()` dialogs, or permissions policy to turn off autoplaying media, those features are [always disabled](https://github.com/WICG/portals#other-restrictions-while-portaled) in pre-activation portals. And because portals are isolated from communicating with their embedder pre-activation, any problems which CSP Embedded Enforcement would attempt to protect against will instead be caught by this communications barrier and prevented from impacting the embedder.
 
+### Activation
+
+The basics of activation are explained [in the intro example](#example): calling `portalElement.activate()` causes the embedding window to navigate to the content which is already loaded into the portal. This section discusses some of the subtler details.
+
+First, note that a portal may be in a "closed" state, when it is not displaying valid, activatable content. This could happen for several reasons:
+
+- The host page author has incorrectly set the portal to a non-HTTP(S) URL, e.g. using `<portal src="data:text/html,hello"></portal>`. Portals can only display HTTP(S) URLs.
+- The portaled page cannot be loaded, for reasons outside of the host page author's control. For example, if the portaled content does a HTTP redirect to a `data:` URL, or if the portaled content gives a network error.
+- The user is offline, which also causes a network error.
+
+(What, exactly, the `<portal>` element displays in this state is still under discussion: [#251](https://github.com/WICG/portals/issues/251).)
+
+Attempting to activate a closed portal will fail. Activation can also fail if another navigation is in progress, as discussed [above](#session-history-navigation-and-bfcache). In all of these cases, the promise returned by the `activate()` method will be rejected, allowing page authors to gracefully handle the failure with a custom error experience.
+
+Another consideration is how activation behaves when the portal is currently loading content. This breaks down into two cases:
+
+- During the initial load of content into a portal, e.g. given
+
+  ```js
+  const portal = document.createElement("portal");
+  portal.src = "https://slow.example.com/";
+  document.body.append(portal);
+  portal.activate();
+  ```
+
+  the promise returned by `activate()` will not settle until the navigation is far enough along to determine whether or not it will be successful. This requires waiting for the response to start arriving, to ensure there are no network errors and that the final response URL is a HTTP(S) URL. Once it reaches that point, then the promise will fulfill or reject appropriately. If the promise fulfills, then activation will have completed, and the content will be loading into the newly-activated browsing context. If it rejects, then no activation will have occurred.
+
+- After the initial load of the portal, e.g. given
+
+  ```js
+  const portal = getSomeExistingFullyLoadedPortal();
+  portal.src = "https://different-url.example.com/";
+  portal.activate();
+  ```
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -300,6 +300,67 @@ An additional reason for avoiding these mechanisms is that it makes writing port
...
+
+- After the initial load of the portal, e.g. given
+
+  ```js
+  const portal = getSomeExistingFullyLoadedPortal();
+  portal.src = "https://different-url.example.com/";
+  portal.activate();
+  ```
+
+  activation of the already-loaded content will happen immediately, and the navigation to the new content will happen in the newly-activated browsing context. In these cases, the promise returned by `activate()` will generally fulfill, as it is almost always possible to activate the already-loaded content. (The exceptions are edge cases like if another user-initiated navigation, or another portal activation, is already ongoing.)
```
jakearchibald, line 338: I still think this bypasses CSP rules, but I guess we can continue debating it in the issues.

README.md
```diff
@@ -300,6 +300,67 @@ An additional reason for avoiding these mechanisms is that it makes writing port
...
+
+First, note that a portal may be in a "closed" state, when it is not displaying valid, activatable content. This could happen for several reasons:
+
+- The host page author has incorrectly set the portal to a non-HTTP(S) URL, e.g. using `<portal src="data:text/html,hello"></portal>`. Portals can only display HTTP(S) URLs.
+- The portaled page cannot be loaded, for reasons outside of the host page author's control. For example, if the portaled content does a HTTP redirect to a `data:` URL, or if the portaled content gives a network error.
+- The user is offline, which also causes a network error.
+
+(What, exactly, the `<portal>` element displays in this state is still under discussion: [#251](https://github.com/WICG/portals/issues/251).)
+
+Attempting to activate a closed portal will fail. Activation can also fail if another navigation is in progress, as discussed [above](#session-history-navigation-and-bfcache). In all of these cases, the promise returned by the `activate()` method will be rejected, allowing page authors to gracefully handle the failure with a custom error experience.
```
jakearchibald, line 315: I'm still a little worried that portal activation rejections can tell you something about a page that you didn't already know, but we can work on that in issues.

jakearchibald (approved):

jeremyroman (approved):
lgtm2

README.md
kjmcnee, line 338: Which context the navigation occurs in depends on how we want to interpret activate in this case.
If activate cancels the ongoing navigation in the portal's new context and restarts it in the activated context at the top level, then yes it happens in the newly-activated browsing context.
If activate promotes both the existing context and the context with the pending navigation to the top level, then the navigation continues in said context. When it matures, we navigate the top level to the new context.

That said, that's probably excessive detail for the purposes of this section. The specifics of this are more likely to affect things like CSP, as Jake mentioned. I might just rephrase this to "the navigation to the new content will happen at the top level."

kjmcnee (approved):
Generally LGTM with a bit of nuance about the browsing contexts involved in the navigation.

README.md
jeremyroman: Hmm, didn't we decide that assigning src made a new browsing context? If so, I'd have expected that browsing context to be used, and the old one to be detached immediately when src is assigned. That is, I would expect `activate` to send you to the last URL assigned to `src` (+ redirects).

jeremyroman:
left a comment in response to Jake's on line 338; GitHub doesn't seem to want to notify?

README.md
domenic: Oh, right, I forgot about that consideration. I do agree it'll be simpler to just discard when src="" is assigned. OK, so the example I'm really thinking of is an in-page navigation.

This will require a reasonably-big update to this PR. And I'll try to update the spec with discard-on-src-change as well.

domenic:
Alright, I've kind of morphed this into a bigger change. Now it includes spec changes, and tackles #197 at the same time. I'd like @jeremyroman's review in particular for spec mechanics. If @kjmcnee has time his re-review would also be appreciated, but I think conceptually there's not too much delta from when he last looked at it; we just incorporated the src=""-changes-cause-reset which had previously been discussed.

jeremyroman:
LGTM. This makes sense to me (obviously we still have to make corresponding impl changes).

kjmcnee:
> Alright, I've kind of morphed this into a bigger change. Now it includes spec changes, and tackles #197 at the same time. I'd like @jeremyroman's review in particular for spec mechanics. If @kjmcnee has time his re-review would also be appreciated, but I think conceptually there's not too much delta from when he last looked at it; we just incorporated the src=""-changes-cause-reset which had previously been discussed.

Still LGTM. Sorry for the delay.

PR #254, opened 2020-10-09
Don't suggest that unload handlers would necessarily be run following activation

kjmcnee:
Currently, a note suggests that closing the predecessor following a lack
of adoption fires unload handlers. This is not necessarily the case
given bfcache. We update this note and reference the issue for
discussing unload handlers.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/portals/pull/254.html" title="Last updated on Oct 9, 2020, 11:29 PM UTC (8327473)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/254/3c57d0d...8327473.html" title="Last updated on Oct 9, 2020, 11:29 PM UTC (8327473)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -28,6 +28,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
             text: create a new top-level browsing context; url: creating-a-new-top-level-browsing-context
         urlPrefix: browsing-the-web.html
             text: prompt to unload; url: prompt-to-unload-a-document
+            text: unload; url: unload-a-document
             text: reserved environment; for: navigation params; url: navigation-params-reserved-environment
             text: request; for: navigation params; url: navigation-params-request
         urlPrefix: common-dom-interfaces.html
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -233,6 +234,20 @@ spec: ecma-262; urlPrefix: http://tc39.github.io/ecma262/
                 The user agent *should not* ask the user for confirmation during the
                 [=prompt to unload=] step (and so the browsing context should be
                 [=discard a browsing context|discarded=]).
```
kjmcnee, line 236: I don't think closing the browsing context is what we want here. Even if we don't ask for confirmation, running beforeunload _after_ the navigation has already finished wouldn't be the right time. If we do want to run beforeunload, it should be before step 4 of activate-a-portal-browsing-context. Also step 3 of closing (remove browsingContext from the user interface) is redundant given step 6 of activation. It seems like we should replace closing with unload and discard. WDYT?

kjmcnee:
Adding @jeremyroman. PTAL.

jeremyroman (approved):
sorry for dropping the ball on this

PR #255, opened 2020-10-21
Rebase portals explainer on prerendering explainers

domenic:
Follows https://github.com/jeremyroman/alternate-loading-modes/pull/5.

The only intentional normative change here is that we now respect `prefetch-src` and use it as a fallback for `portal-src`.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/portals/pull/255.html" title="Last updated on Oct 30, 2020, 4:04 PM UTC (de6f119)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/portals/255/3c57d0d...de6f119.html" title="Last updated on Oct 30, 2020, 4:04 PM UTC (de6f119)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ for (const link of document.querySelectorAll('a.seamless')) {
     // Once the preview is now displayed as the whole viewport, activate.
     // This performs the instant navigation/URL bar update/etc.
     try {
-      portal.activate();
+      await portal.activate();
     } catch {
       // If activation failed, restore the portal to hidden (so that back-navigations
       // don't show the full-viewport portal), and fall back to a normal navigation.
@@ -176,14 +176,14 @@ An embedder which portals same-origin content has the ability to communicate wit
 
 As shown in the [introductory example](#introductory-example), this is done by exposing the `postMessage()` method directly on the `HTMLPortalElement` interface. Unlike an iframe, there is no direct access to the `contentWindow` of the portaled content; message passing is the only interface available.
 
-Additionally, all pages get a `window.portalHost` property, which is non-null for same-origin portaled content. This is the way that portaled content recieves or transmits messages from or to its embedder:
+Additionally, all pages get a `window.portalHost` property, which is non-null for all portaled content. This is the way that portaled content recieves or transmits messages from or to its embedder:
 
 ```js
 window.portalHost.onmessage = e => { /* ... */ };
 window.portalHost.postMessage(/* ... */);
 ```
 
-Cross-origin portaled content cannot communicate with its embedder, as doing so would [allow cross-site tracking](https://github.com/jeremyroman/alternate-loading-modes/blob/gh-pages/browsing-context.md#privacy-based-restrictions).
+Cross-origin portaled content cannot communicate with its embedder, as doing so would [allow cross-site tracking](https://github.com/jeremyroman/alternate-loading-modes/blob/gh-pages/browsing-context.md#privacy-based-restrictions). In those cases, no `message` events will be sent to `window.portalHost`, and `window.portalHost.postMessage()` will not deliver any message.
 
 ### Rendering
 
--- a/README.md
+++ b/README.md
@@ -114,7 +114,12 @@ for (const link of document.querySelectorAll('a.seamless')) {
 
     // Once the preview is now displayed as the whole viewport, activate.
     // This performs the instant navigation/URL bar update/etc.
-    portal.activate();
+    try {
+      portal.activate();
+    } catch {
+      // If activation failed, fall back to a normal navigation.
+      location.href = link.href;
+    }
   };
 }
 </script>
@@ -207,7 +212,7 @@ Authors should respect the `prefers-reduced-motion` media query by conditionally
 
 ### Session history, navigation, and bfcache
 
-At a base level, portals behave the same as other prerendering browsing contexts do with respect to [session history](https://github.com/jeremyroman/alternate-loading-modes/blob/gh-pages/browsing-context.md#session-history) and [navigation](https://github.com/jeremyroman/alternate-loading-modes/blob/gh-pages/browsing-context.md#navigation). To summarize, content inside the portal has a trivial session history, and activation acts like a navigation of the host page, appending the current document in the portal to the host page's session history. This works to preserve user expectations for the back button. Note that this is very different from how iframes behave, and is one of the reasons it is better to think of portals as "inline-displayed popups" or "prerendered links" than as iframes. (Discussed further [below](#summary-of-differences-between-portals-and-iframes).)
+At a base level, portals behave the same as other prerendering browsing contexts do with respect to [session history](https://github.com/jeremyroman/alternate-loading-modes/blob/gh-pages/browsing-context.md#session-history) and [navigation](https://github.com/jeremyroman/alternate-loading-modes/blob/gh-pages/browsing-context.md#navigation). To summarize, content inside the portal has a trivial session history, and activation acts like a navigation of the host page, appending the portal's current session history entry to the host page's session history. This works to preserve user expectations for the back button. Note that this is very different from how iframes behave, and is one of the reasons it is better to think of portals as "inline-displayed popups" or "prerendered links" than as iframes. (Discussed further [below](#summary-of-differences-between-portals-and-iframes).)
 
 Because of the predecessor adoption feature, portals have some additional complexity, where they can cause the predecessor (i.e., the host document) to move into a prerendering browsing context. TODO describe this more.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -70,21 +77,60 @@ window.addEventListener('portalactivate', e => {
...
+
+    // Show the portal, and animate it to the whole viewport over 300 milliseconds.
+    portal.hidden = false;
+    if (!matchMedia('(prefers-reduced-motion: reduce)').matches) {
+      await portal.animate([{ width: '100vw', height: '100vh' }], { duration: 300 }).finished;
+    }
+
+    // Once the preview is now displayed as the whole viewport, activate.
+    // This performs the instant navigation/URL bar update/etc.
+    portal.activate();
```
kjmcnee, line 117: Optional nit: Might be good to illustrate error handling, at least at a high level. Something like:

```
try {
  await portal.activate()
} catch (e) {
  // Conflicting navigation, we should hide the portal.
}
```

README.md
```diff
@@ -110,209 +158,84 @@ Non-goals:
...
+At a base level, portals behave the same as other prerendering browsing contexts do with respect to [session history](https://github.com/jeremyroman/alternate-loading-modes/blob/gh-pages/browsing-context.md#session-history) and [navigation](https://github.com/jeremyroman/alternate-loading-modes/blob/gh-pages/browsing-context.md#navigation). To summarize, content inside the portal has a trivial session history, and activation acts like a navigation of the host page, appending the current document in the portal to the host page's session history. This works to preserve user expectations for the back button. Note that this is very different from how iframes behave, and is one of the reasons it is better to think of portals as "inline-displayed popups" or "prerendered links" than as iframes. (Discussed further [below](#summary-of-differences-between-portals-and-iframes).)
```
kjmcnee, line 210: nit: "appending the current document in the portal" ->  "appending the portal's current session history entry"

README.md
domenic: Good point. Maybe I can eliminate the closed check even.

README.md
```diff
@@ -110,209 +158,84 @@ Non-goals:
...
 
 ### Activation
 
-The basics of activation are explained [in the intro example](#example): calling `portalElement.activate()` causes the embedding window to navigate to the content which is already loaded into the portal. This section discusses some of the subtler details.
+The basics of activation are explained [in the intro examples](#examples): calling `portalElement.activate()` causes the embedding window to navigate to the content which is already loaded into the portal, i.e. it is a developer-controlled way of activating the general activation operation that all prerendering browsing contexts have. This section discusses some of the subtler details created by exposing this functionality to developers, instead of leaving it up to the browser as other prerendering browsing contexts do.
```
kjmcnee, line 224: nit: "way of activating the general activation operation" -> "way of performing the general activation operation" ?

README.md
kjmcnee: The check could still be useful if the page wants to avoid a useless animation of a portal that has no chance of successfully activating. For pages that don't noticeably change before activation (e.g. offscreen prerendering), then yeah, EAFP.

README.md
```diff
@@ -70,21 +77,65 @@ window.addEventListener('portalactivate', e => {
...
+      await portal.animate([{ width: '100vw', height: '100vh' }], { duration: 300 }).finished;
+    }
+
+    // Once the preview is now displayed as the whole viewport, activate.
+    // This performs the instant navigation/URL bar update/etc.
+    try {
+      portal.activate();
+    } catch {
+      // If activation failed, fall back to a normal navigation.
+      location.href = link.href;
```
kjmcnee, line 121: If the reason activation failed is that the user caused a navigation, then this would likely also be ignored. So, it'd be good to return the page to a usable state as well.

For example, consider the following sequence:
User presses back and the back navigation is slow to load, the portal activates and is rejected, the page attempts to navigate and the navigation is ignored, then the user presses stop to stay on the current page. We're left with a portal filling the page.

domenic: Done, thanks again!

README.md
```diff
@@ -70,21 +77,68 @@ window.addEventListener('portalactivate', e => {
...
+    // Show the portal, and animate it to the whole viewport over 300 milliseconds.
+    portal.hidden = false;
+    if (!matchMedia('(prefers-reduced-motion: reduce)').matches) {
+      await portal.animate([{ width: '100vw', height: '100vh' }], { duration: 300 }).finished;
+    }
+
+    // Once the preview is now displayed as the whole viewport, activate.
+    // This performs the instant navigation/URL bar update/etc.
+    try {
+      portal.activate();
```
kjmcnee, line 118: I think we're missing an 'await' here.

README.md
```diff
@@ -70,21 +77,68 @@ window.addEventListener('portalactivate', e => {
...
+      return;
+    }
+
+    e.preventDefault();
+
+    // Show the portal, and animate it to the whole viewport over 300 milliseconds.
+    portal.hidden = false;
+    if (!matchMedia('(prefers-reduced-motion: reduce)').matches) {
+      await portal.animate([{ width: '100vw', height: '100vh' }], { duration: 300 }).finished;
+    }
```
kjmcnee, line 113: Sorry, one more thing I just noticed. If we have prefers-reduced-motion, it looks like this would flash the portal with its preview size. I think we would want something like:
```
if (!matchMedia('(prefers-reduced-motion: reduce)').matches) {
  portal.hidden = false;
  await portal.animate([{ width: '100vw', height: '100vh' }], { duration: 300 }).finished;
} else {
  portal.style.width = '100vw';
  portal.style.height = '100vh';
  portal.hidden = false;
}
```

domenic: I assumed activate would be instant (i.e. before the next animation frame), so you wouldn't see the preview. Maybe that's not true in practice?

kjmcnee: Not sure off the top of my head. In the case where the successor is slow to produce a frame even after activation, we would temporarily show the old content. That might be relevant here.

I was thinking more of the case where the user clicks the link right away so that the activate call below wouldn't be completed immediately while waiting for the initial nav.

And upon further reflection, restyling the preview for the prefers-reduced-motion case is probably unnecessary anyway. We could just activate when ready. That is, move the |portal.hidden = false;| into the if block, so this just behaves like offscreen prerender.

domenic: I was under the impression that we'd make the portal BC full-viewport and change the URL bar immediately, even if the content was still loading. (Even for the initial nav.)

But yeah, +1 for staying hidden in the prefers-reduced-motion case.

kjmcnee: Yes, once the navigation matures, the URL bar would immediately reflect the portal context as it still loads. It's just that if the portal context has nothing to show yet, we would keep showing the predecessor's frame (as in image) for a short time to avoid the image flashing. The same thing is done with regular navigations.

README.md
```diff
@@ -110,209 +167,84 @@ Non-goals:
...
-  - [Page Visibility](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) and the timing of `requestAnimationFrame()` callbacks match the visibility of the top-level page, as in iframes, to prevent the page from encoding messages in visibility changes. However, this leads to the possibility that a portal and containing page could use the timing of user-caused visibility changes to join the user across site boundaries. Whether and how to prevent this is [still under discussion](https://github.com/WICG/portals/issues/193#issuecomment-639768218).
+As shown in the [introductory example](#introductory-example), this is done by exposing the `postMessage()` method directly on the `HTMLPortalElement` interface. Unlike an iframe, there is no direct access to the `contentWindow` of the portaled content; message passing is the only interface available.
 
-#### Communications channels that match navigation
+Additionally, all pages get a `window.portalHost` property, which is non-null for all portaled content. This is the way that portaled content recieves or transmits messages from or to its embedder:
```
jeremyroman, line 180: sp: receives

README.md
```diff
@@ -110,209 +167,84 @@ Non-goals:
...
-
-- Popups, pointer lock, orientation lock, the presentation API, downloads, and modal dialogs (`alert()` etc.) all are disabled pre-activation. (These are features which are currently only possible to disable with through [iframe sandboxing](https://html.spec.whatwg.org/multipage/origin.html#sandboxing).)
-
-After activation, these restrictions are lifted: the portaled content is treated like a normal top-level browsing context, and is able to use all these features in the normal way. The `portalactivate` event can be used to request permissions upon activation, if necessary. (Although doing so gives a user experience equivalent to requesting permissions on load, and thus is rarely the best design.)
-
-_Note: ideally Permissions Policy would become a superset of the Permissions API and iframe sandboxing. Then we could use that infrastructure as the single place to impose and lift these restrictions. Currently that is not the case, so the spec will be more messy._
+```js
+window.portalHost.onmessage = e => { /* ... */ };
```
jeremyroman, line 183: nit: maybe we should ideally advise `addEventListener` to avoid clobbering, though I guess this issue is pretty well understood among web devs already

README.md
```diff
@@ -70,21 +77,69 @@ window.addEventListener('portalactivate', e => {
...
+      // Let the normal link click go through.
+      return;
+    }
+
+    e.preventDefault();
+
+    // Show the portal, and animate it to the whole viewport over 300 milliseconds.
+    if (!matchMedia('(prefers-reduced-motion: reduce)').matches) {
+      portal.hidden = false;
+      await portal.animate([{ width: '100vw', height: '100vh' }], { duration: 300 }).finished;
```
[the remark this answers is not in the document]
domenic: I'm not sure, what does that mean?

domenic:
Alright, I'll merge this, especially so that #256 can have a stable base to work with. Thanks everyone for the review, and especially the iteration on the example! I imagine there's still room for improvement, which we can continue in future PRs...

PR #258, opened 2020-11-17
Fix README links to alternate-loading-mode

bokand:
The gh-pages branch no longer has has any of this content so the links lead to 404


domenic (approved):
Thank you!!

PR #291, opened 2024-07-20
Publish a notice that WICG/portals is inactive

jeremyroman:
This should make it clearer for anyone who comes by why there hasn't been recent activity, and points in some directions they could look to see where this work is continuing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,15 @@
+The Portals proposal is not currently being actively pursued. 
+
+Many of the capabilities that Portals would have enabled are being explored in other proposals:
+* [navigation speculation](https://github.com/WICG/nav-speculation) (including speculation rules, prefetch, and prerendering)
+* [View Transitions](https://drafts.csswg.org/css-view-transitions-1/)
+* [Fenced Frame](https://wicg.github.io/fenced-frame/)
+* and more (which may be added here)
+
+This repository is thus effectively archived for the time being.
+
+---
+
 # Portals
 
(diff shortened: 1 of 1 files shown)


jeremyroman:
@a4sriniv @domenic Make sense? Adithya suggested updating this repository after the related Chromium code was removed. Especially if there's anything else that should be included in this list.

domenic:
Yeah, this makes sense to me. I wonder if we want to actually archive it?

jeremyroman:
Maybe -- I don't know what that process entails or what its implications are. Do you happen to?

(it would be nice if it were still available and not a 404, for instance)

lucasgadani:
LGTM

domenic:
https://docs.github.com/en/repositories/archiving-a-github-repository/archiving-repositories . Note that you can always un-archive.

README.md
```diff
@@ -1,3 +1,15 @@
+The Portals proposal is not currently being actively pursued. 
+
+Many of the capabilities that Portals would have enabled are being explored in other proposals:
+* [navigation speculation](https://github.com/WICG/nav-speculation) (including speculation rules, prefetch, and prerendering)
```
a4sriniv, line 4: nit: should we link to https://wicg.github.io/nav-speculation/ for consistency with the fenced frame link below?

LGTM otherwise!

jeremyroman: That page is mostly an index, and I think the README.md gives a better overview of the work.

